Issue #452, opened 2017-08-25
Unified Resource Handling and New Magic Wand Syntax


viper-admin:
>  **Pull request** :twisted_rightwards_arrows: created by bitbucket user **robinsierra** on 2017-08-25 20:19
> Last updated on 2017-09-02 20:09
> Original Bitbucket pull request id: 42
>
> Participants:
>
> * **@mschwerhoff** (reviewer)
> * **@vakaras** (reviewer)
> * bitbucket user **robinsierra**
> * **@alexanderjsummers** (reviewer)
>
> Source: https://github.com/viperproject/silicon/commit/8324926bec264e73ec652a995f91347b57d1ff9d on branch `robinsierra/silicon_qr/default`
> Destination: https://github.com/viperproject/silicon/commit/2171ace54c63e2f46966ae1bb32dda8439f9596f on branch `master`
> Marge commit: https://github.com/viperproject/silicon/commit/8324926bec264e73ec652a995f91347b57d1ff9d
>
> State: **`MERGED`**

**Nils:**

* The new syntax for packaging magic wands looks like:
 ```
package A --* B { /* statements as ghost operations */}
```
* ```lhs```-expressions have been replaced by ```old[lhs]```-expressions
* magic wand variables have been removed
* ```applying``` can be used analogously to ```unfolding``` to temporarily apply a wand without side-effects
* ```assert``` can be used to add permissions to the footprint of a magic wand during package operations
* quantified magic wands and the usage of quantified permissions in magic wands and proof scripts is now supported

**Robin:**

* Resource descriptions have been separated from non-quantified chunks: Properties like e.g. `c.perm <= 1` or `forEach c1, c2 :: c1.perm + c2.perm > 1 ==> c1.args != c2.args` for fields can now be specified in a resource description
* FieldChunk and PredicateChunk are unified
* Producing is unified for all non-quantified chunks
* An additional option for using a more complete exhaling has been added
* Partial extension of resouce description to work for quantified chunks
* Static optimization for general permission literals
* Fixed issues <https://github.com/viperproject/silicon/issues/48> and <https://github.com/viperproject/silicon/issues/217>.

PR #492, opened 2020-05-02
A workaround for test argument passing

maurobringolf:
Passing test arguments through scalatest, i.e. `testOnly -- -Dsilicon:logLevel=DEBUG` is not working for me. I could not figure out the precise root issue, but found a small workaround. Maybe someone else can understand and fix the root issue using this?

@mschwerhoff 

https://github.com/viperproject/silver/pull/452

### DIFF
--- a/src/test/scala/SiliconTests.scala
+++ b/src/test/scala/SiliconTests.scala
@@ -56,15 +56,16 @@ class SiliconTests extends SilSuite {
     }
   }
 
-  lazy val verifiers = List(createSiliconInstance())
+  lazy val verifiers = verifiers(Map.empty)
+  override def verifiers(configMap: Map[String, Any]) = List(createSiliconInstance(configMap))
 
   val commandLineArguments: Seq[String] =
     Seq("--timeout", "300" /* seconds */)
 
-  private def createSiliconInstance() = {
+  private def createSiliconInstance(configMap: Map[String, Any]) = {
     val args =
       commandLineArguments ++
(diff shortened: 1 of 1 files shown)


mschwerhoff:
@maurobringolf The workaround looks OK to me. I remember that getting argument forwarding to work took quite some time and was harder than expected ...

maurobringolf:
Ah okay, then it is probably easier to use this "manual" argument passing. I will notify you once I cleaned up both PRs and tested them against the other test suites.

maurobringolf:
Okay so my initial proposal did not make a lot of sense, because multiple calls to `verifiers(configMap)` multiple times would create multiple verifier instances. Instead we need access to the very same instances in order to `start()` and `stop()` them from the test suite. So I did not change the `def verifier()` interface, but added another method `configureVerifiersFromConfigMap(configMap)` which is called just before the verifiers are started. A testsuite that needs to access the `configMap` can implement this method and configure its verifiers accordingly.

If this behaves as intended, then existing test suites are not affected since they will just configure their verifiers in `def verifier() / lazy val verifiers` and use the default empty `configureVerifiersFromConfigMap` method.

I tested this in Silicon (`sbt test`), Carbon ( `sbt test`) and also am able to run `benchmark.sh` without problems. So from my perspective it is good to go.

mschwerhoff:
Just to double-check: PRs #492 (Silicon) and [#452](https://github.com/viperproject/silver/pull/452)  (Silver) can be merged in, right?

maurobringolf:
Yes, all local tests succeed.

PR #506, opened 2020-07-27
Add type parameters to SMT quantifier ids of domain axioms (helfpul for smt.qi.profile)

maurobringolf:
While looking at the `smt.qi.profile` of some Silicon logfiles I found duplicate quantifier IDs because the Viper program created multiple instances of `Set[]`. In my example only one of them was expensive, so I needed to tell them apart. For this, I added the (translated) type parameter instantiation to the quantifier ids of domain axioms and I think that could be helpful in other cases too.

I am not familiar with the code that instantiates domains and my attempt at adding this information feels a bit clumsy, so I am open to suggestions.


src/main/scala/state/Terms.scala
```diff
@@ -521,7 +521,7 @@ object Quantification
   def apply(q: Quantifier, vars: Seq[Var], tBody: Term, triggers: Seq[Trigger], name: String)
            : Quantification = {
 
-    apply(q, vars, tBody, triggers, "", false)
+    apply(q, vars, tBody, triggers, name, false)
```
maurobringolf, line 524: This is an unrelated small bug that ignores given quantifier names.

PR #528, opened 2021-01-06
Counterexample API

Aurel300:
Description (by @cedihegi):

> To use these new extracted counterexamples, you can use the flag `--counterexample mapped`. The new trait `ExtractedModelEntry` should make the counterexamples more readable and extract information from the heap, but heap values such as predicates can still be accessed, also in a more processed way since all the first/second/Sortwrappers should be gone now and replaced with their corresponding values in the extracted Heap fields of the converter class.
> A lot of types like sequences etc are not handled yet, and I am not sure if I plan to do so since they are not used in Prusti.
> To see the advantage of these new counterexamples you can run it on the files in `src/test/resourcers/demo-counterexamples`, the file `simple-refs-rec.vpr` is probably the most interesting.


src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+            case Combine(p0, p1) => {
+                val p0eval = evaluateTerm(p0, model)
+                val p1eval = evaluateTerm(p1, model)
+                return "($Snap.combine " + p0eval + " " + p1eval + ")"
+            }
+            case First(p) => {
+                val sub = evaluateTerm(p, model)
+                if (sub.startsWith("($Snap.combine")){
+                    return getParts(sub)(1)
+                }
```
Aurel300, line 167: I guess this is still inherited from Nagini, but it would be much better if there was no stringification followed by parsing. Should `evaluateTerm` maybe return `ExtractedModelEntry` values? Maybe it doesn't yet have all the variants it needs to have?

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+
+/* Explanation: (to be removed later)
+    To use these new extracted counterexamples, you can use the flag "--counterexample mapped"
+    The new trait ExtractedModelEntry should make the counterexamples more readable and extract
+    information from the heap, but heap values such as predicates can still be accessed,
+    also in a more processed way since all the first/second/Sortwrappers should be gone now and 
+    replaced with their corresponding values in the extracted Heap fields of the converter class
+
+    A lot of types like sequences etc are not handled yet, and I am not sure if I plan to do so
+    since they are not used in Rust
```
Aurel300, line 23: It might be difficult for you to test, but we do have plans for more Viper types exposed to Prusti, see https://github.com/viperproject/prusti-dev/pull/181

fpoli:
@cedihegi Do you plan to clean up a bit the commit history before merging? We don't have strict rules, but ".." and "changes?" are a bit too much :D

([This tutorial](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) explains how to rewrite the commit history, while [this article](https://chris.beams.io/posts/git-commit/) explains what an ideal commit message should look like.)

cedihegi:
@fpoli Yes I am sorry for this mess, thanks for the resources!
Is it okay as of now?

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+
+    To see the advantage of these new counterexamples you can run it on the
+    files in src/test/resourcers/demo-counterexamples, the file simple-refs-rec.vpr is 
+    probably the most interesting.
+
+*/
+
+sealed trait ExtractedModelEntry
+case class LitIntEntry(value: BigInt) extends ExtractedModelEntry {
+    override def toString = value.toString
```
mschwerhoff, line 36: Please stick to the coding style that is typically used for Silicon (and for to some extend Silver).

Here: indent with two spaces

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+}
+case class LitBoolEntry(value: Boolean) extends ExtractedModelEntry {
+    override def toString = value.toString
+}
+case class RefEntry(fields: Map[String, ExtractedModelEntry]) extends ExtractedModelEntry {
+    override def toString : String = {
```
mschwerhoff, line 42: Style: types as `e: T`, not `e : T` or `e:T` etc.

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+        var buf = "Ref {\n"
+        for ((name, entry) <- fields) {
+            buf += "\t" + name + " <- " + entry.toString.split("\n").mkString("\n\t") + "\n"
+        }
+        buf += "}"
+        return buf
```
mschwerhoff, line 48: Style: avoid `return`. While [actually harmful](https://tpolecat.github.io/2014/05/09/return.html) only in lambdas, it's considered bad style in Scala to use them.

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+    }
+}
+case class VarEntry(name: String) {
+    override def toString = name
+}
+case class NullRefEntry() extends ExtractedModelEntry {
+    override def toString = "Null()"
+}
+case class OtherEntry(value: String) extends ExtractedModelEntry {
+    override def toString = value + "(unhandled type)"
```
mschwerhoff, line 58: Style: prefer string interpolation over concatenation, e.g. `s"$value (unhandled type)"` (or `s"${value}(unhandled type)"`, if you really don't want a space here)

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+}
+
+case class ExtractedModel(entries: Map[String, ExtractedModelEntry]) {
+    override def toString : String = {
+        var buf = ""
+        for ((name, entry) <- entries) {
```
mschwerhoff, line 64: Style: check out function `mkString`, or consider using a `foldLeft`; both should help you avoiding the mutable `buf` and the loop

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+    override def toString = recv + "." + field
+}
+
+case class ExtractedHeap(entries: Map[HeapEntry, String]) {
+    override def toString = entries.map(x => x._1.toString + " <- " + x._2).mkString("\n")
+}
+
+
+/* basically a 1 to 1 copy of nagini code */
+object Converter{
```
mschwerhoff, line 85: Not that Silicon strictly follows best-practice software engineering, but it seems odd that so much code would reside in a file in the `interfaces` directory

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+        return res
+    }
+
+    def getFunctionValue(model: Model, fname: String, args: String) : String = {
+        val entry : ModelEntry = model.entries(fname)
+        entry match {
+            case SingleEntry(s) => return s 
+            case MapEntry(m: Map[Seq[String], String], els:String) =>
+                val filtered = m.filter(x => snapToOneLine(x._1.mkString(" ")) == args)
+                if (filtered.size >= 1) {
```
mschwerhoff, line 103: `filtered.nonEmpty` is potentially more performant than `filtered.size >= 1`

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    def sortToType(s:Sort) : Option[ast.Type] = {
+        s match {
+            case sorts.Ref => return Some(ast.Ref)
+            case sorts.Int => return Some(ast.Int)
+            case sorts.Bool => return Some(ast.Bool)
+            case sorts.Perm => return Some(ast.Perm)
+            case _ => println("This type of Heap Field can not be handled : " + s.toString)
```
mschwerhoff, line 128: `println`ing an error is most likely not a good idea. If you want debug output, use Silicon's logger(s) for that. If you want an error, raise an exception.

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+            case IntLiteral(n) => return term.toString
+            case t:BooleanLiteral => return t.value.toString
+            case Null() => return model.entries("$Ref.null").toString
+            case Var(id, sort) => {
+                val key = term.toString
+                //this can fail : TODO throw and catch exception
+                return model.entries(key).toString
+            }
+            case t@App(app, args) => {
+                println("Found type APP")
```
mschwerhoff, line 146: Debug output?

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+                if (!model.entries.contains(fname)){
+                    fname = app.id.toString
+                    if (!model.entries.contains(fname)){
+                        fname = fname.replace("[","<").replace("]",">")
+                    }
+                }
+                var arg = snapToOneLine(args.map(x => evaluateTerm(x, model)).mkString(" "))
+                return getFunctionValue(model, fname, arg)
+            }
+            case Combine(p0, p1) => {
```
mschwerhoff, line 158: `case`s don't need braces

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    def extractHeap(h:Iterable[Chunk], model: Model) : ExtractedHeap = {
+        var target : ExtractedHeap = Map()
+        for (chunk <- h) {
+            chunk match {
+                case c@BasicChunk(resId, id, args, snap, perm) => {
+                    resId match {
+                        case FieldID => val (entry, value) = extractField(c, model)
```
mschwerhoff, line 206: Style: don't start multiline blocks right after `=>`

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+                return getFunctionValue(model, lookupFuncName, arg)
+            }
+            case _ =>   println("of unhandled type") 
+                        return ""
+        }
+    }
+
+    def extractHeap(h:Iterable[Chunk], model: Model) : ExtractedHeap = {
+        var target : ExtractedHeap = Map()
+        for (chunk <- h) {
```
mschwerhoff, line 202: Style: lines 202 to 204 should be equivalent to

    h foreach {
      case c@BasicChunk ...

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+                    
+                }
+                case _ => println("WARNING: not handling non-basic chunks")
+            }
+        }
+        return target
+    }
+
+    def extractField(chunk:BasicChunk, model: Model) : (HeapEntry, String) =  {
+        val fieldname = chunk.id.name //String
```
mschwerhoff, line 222: Types as code comments?

src/main/scala/interfaces/Converter.scala
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    def extractedModelToModel(extModel: ExtractedModel) : Model = {
+        val map = extModel.entries.map(x => (x._1 -> SingleEntry(x._2.toString)))
+        return Model(map)
+    }
+}   
+    
+
+case class Converter(model: Model, store: Store, heap: Iterable[Chunk], oldHeaps: State.OldHeaps){
```
mschwerhoff, line 310: Style: space before braces

src/main/scala/interfaces/Verification.scala
```diff
@@ -113,3 +114,18 @@ case class SiliconVariableCounterexample(internalStore: Store, nativeModel: Mode
...
+case class SiliconMappedCounterexample(internalStore: Store, heap: Iterable[Chunk], oldHeaps: State.OldHeaps, nativemodel: Model) extends SiliconCounterexample {
+  val converter = Converter(nativemodel, internalStore, heap, oldHeaps)
+  
+  override def toString = {
+    var buf : String = ""
+    buf += converter.modelAtLabel.map(x => "model at label: " + x._1 + "\n" + x._2.toString).mkString("\n")
+    buf += "\nfinally: \n" + converter.extractedModel.toString
+    buf
+  }
+  val model : Model = nativemodel
```
mschwerhoff, line 126: Style: separate defs and val with newlines

src/main/scala/rules/SymbolicExecutionRules.scala
```diff
@@ -29,14 +29,20 @@ trait SymbolicExecutionRules extends Immutable {
...
       if (model != null && !model.contains("model is not available")){
         val nativeModel = Model(model)
-        var ce: Counterexample = if (Verifier.config.counterexample.toOption.get == "native") {
-          val oldHeap = if (s.oldHeaps.contains(Verifier.PRE_STATE_LABEL))
-            Some(s.oldHeaps(Verifier.PRE_STATE_LABEL).values)
-          else None
-          SiliconNativeCounterexample(s.g, s.h.values, oldHeap, nativeModel)
-        }else{
-          SiliconVariableCounterexample(s.g, nativeModel)
+        val ce_type = Verifier.config.counterexample.toOption.get
```
mschwerhoff, line 32: I believe `config.counterexample()` is recommended over `config.counterexample.toOption.get`

src/main/scala/rules/SymbolicExecutionRules.scala
```diff
@@ -29,14 +29,20 @@ trait SymbolicExecutionRules extends Immutable {
...
+        var ce: Counterexample = ce_type match {
+          case "native" => 
+            val oldHeap = if (s.oldHeaps.contains(Verifier.PRE_STATE_LABEL))
```
mschwerhoff, line 35: `contains` followed by getting the contained value is potentially inefficient (though the difference is most likely irrelevant here).

Better: Call `getOrElse` on `s.oldHeaps`, or get an `Option` from `s.oldHeaps` and use its `fold`

mschwerhoff (changes requested):
I've left lots of small coding style comments. Please check all sources for similar code.

fpoli:
> @fpoli Yes I am sorry for this mess, thanks for the resources!
> Is it okay as of now?

Yes, thanks! :)

src/main/scala/interfaces/Converter.scala
cedihegi: The problem is that the values stored in the model are only available as Strings, so those have to be parsed anyways and it was easier to just do the same with the values obtained from the heap. 
But it should still be possible to do the parsing of model values right in the evaluateTerm function, their types should be available. I will look into this again.

src/main/scala/interfaces/Converter.scala
cedihegi: First of all thanks for the extensive feedback!
I am aware that my current structure does not make a lot of sense, but I thought it would be a good idea to maybe ask you first how you prefer it to be done. Should I put Converter.scala into the reporter package? Or keep the ExtractedHeap and ExtractedModel structure in interfaces and move the rest? I am not very familiar with most of the project.

src/main/scala/interfaces/Converter.scala
cedihegi: I am a bit confused by "silicon's logger". Is it okay if I use the logger from com.typesafe.scalalogging.Logger? I saw it's used in other Files too.
Or what exactly should I use?

src/main/scala/rules/SymbolicExecutionRules.scala
cedihegi: I tried this but can't get it to work because of some typing problems. But this is not really my code anyways, it's still from the original native counterexamples so I hope it's okay if I leave it as it is.

src/main/scala/interfaces/Converter.scala
marcoeilers: FYI, @dohrau changed the model parser yesterday (https://github.com/viperproject/silver/commit/46f31079da934bd884f1d5aaf63d0c378d2ab2c6) so that values are no longer just strings but instances of a new ValueEntry class (that contain strings). Entries that contain function applications, like the ``$Snap.combine`` ones, are now parsed as instances of the ApplicationEntry class.

src/main/scala/interfaces/Converter.scala
mschwerhoff: You're welcome! And no worries, I'm aware that this PR is work in progress. Plus, Silicon's structure & architecture are also not totally thought through. For examples, the `interfaces` package is basically an artefact from my Java background, when I began with Silicon; I don't think it's a common separation to have in Scala.

That said: I suggest to keep everything in package `reporter`, potentially in a subpackage thereof, if you add several files.

src/main/scala/rules/SymbolicExecutionRules.scala
mschwerhoff: Please try again; the original source of the code doesn't really matter. Let me know if you can't get it to work, then I'll have a go at it.

cedihegi: Okay I worked it out using Option.empty instead of None.

src/main/scala/rules/SymbolicExecutionRules.scala
```diff
@@ -32,9 +33,8 @@ trait SymbolicExecutionRules extends Immutable {
         val ce_type = Verifier.config.counterexample()
         var ce: Counterexample = ce_type match {
           case "native" => 
-            val oldHeap = if (s.oldHeaps.contains(Verifier.PRE_STATE_LABEL))
-              Some(s.oldHeaps(Verifier.PRE_STATE_LABEL).values)
-            else None
+            val optOldHeap: Option[Heap] = s.oldHeaps.get(Verifier.PRE_STATE_LABEL)
```
mschwerhoff, line 36: Can't you replace lines 36 and 37 with `s.oldHeaps.get(Verifier.PRE_STATE_LABEL).map(_.values)`?

Do you use IntelliJ? It's pretty good at recommending such simplifications, at least for chained expressions.

cedihegi: I am currently using VSCode but I guess I will check out IntelliJ then.
Thanks for the suggestion, it does work. I was not aware that map works like that on Option instances.

src/main/scala/interfaces/Converter.scala
cedihegi: @marcoeilers @dohrau I adjusted my code to these new entries but while debugging I noticed that negative integers are stored in an ApplicationEntry with name "-" and the positive integer as an argument. Is this actually intended?

dohrau: Yes. This is due to the [smt-lib](http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf) representing negative numbers as "(- n)". The application entry simply reflects this structure as the parser does not perform any transformations or simplifications.

If you want to have integer values, you neet to convert the string values to integers at some point anyway. I suggest you also handle the negation there.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,511 @@
...
+    since they are not used in Rust
+
+    The various parts in this file should probably be moved out of the interfaces package,
+    but for now to keep everything in one place is simpler to work on it.
+
+    To see the advantage of these new counterexamples you can run it on the
+    files in src/test/resourcers/demo-counterexamples, the file simple-refs-rec.vpr is
+    probably the most interesting.
+
+ */
```
Aurel300, line 51: To be removed? I guess this is more of a PR description.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,511 @@
...
+    }
+  }
+
+  def getConstantEntry(s: Sort, m: ModelEntry): ExtractedModelEntry = {
+    s match {
+      case sorts.Ref => VarEntry(m.toString, sorts.Ref)
+      case sorts.Int =>
+        m match {
+          case ConstantEntry(x)             => LitIntEntry(BigInt(x))
+          case ApplicationEntry(name, args) =>
```
Aurel300, line 164: Can this match not be done as `ApplicationEntry("-", [LitIntEntry(n)]) => LitIntEntry(-n)` (or similar?) Is `negate` used anywhere else?

src/main/scala/reporting/Converter.scala
cedihegi: Oh yes, sure. I also just removed all println's.

src/main/scala/reporting/Converter.scala
cedihegi: You are right, this is not really necessary and no, negate is not used anywhere else.

src/main/scala/interfaces/Verification.scala
Aurel300: I think this comment might still apply in the current version @cedihegi

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,531 @@
...
+        getConstantEntry(toSort, m.getOrElse(args, els))
+      case Some(m) => getConstantEntry(toSort, m)
+      case None    => OtherEntry("${fname}", "function not found in model")
+    }
+  }
+
+  def translateSort(s: Sort): String = {
+    s match {
+      case sorts.Set(els: Sort) => s"Set<${translateSort(els)}>"
+      case sorts.Ref            => "$Ref"
```
mschwerhoff, line 170: It would be more stable w.r.t to maintenance if you didn't use a string literal here (and in several other places), which may get out of sync with the rest of Silicon. Not sure how practical it is, but you could potentially use [`TermToSMTLib2Converter`](https://github.com/viperproject/silicon/blob/master/src/main/scala/decider/TermToSMTLib2Converter.scala#L36) to go from `sorts.Ref` to `"$Ref"`.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,531 @@
...
+          .map(t => evaluateTerm(t, model))
+
+        val argsFinal = argEntries.map{
+          case UnprocessedModelEntry(entry) => entry
+          case e: ExtractedModelEntry => ConstantEntry(e.toString)
+        }
+        getFunctionValue(model, fname, argsFinal, toSort)
+
+      case Combine(p0, p1) =>
+        //assuming Combine can only contain other snap.combine and snap.unit
```
mschwerhoff, line 256: That assumption is definitely wrong

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,531 @@
...
+            }
+          case OtherEntry(t, _) => OtherEntry(s"Second($t)", "unapplicable")
+          case _                => OtherEntry(s"Second($sub)", "unapplicable")
+        }
+
+      case SortWrapper(t, to) =>
+        val sub = evaluateTerm(t, model)
+        val fromSortName: String = translateSort(t.sort)
+        val toSortName: String = translateSort(to)
+        val fname = s"$$SortWrappers.${fromSortName}To$toSortName"
```
mschwerhoff, line 292: Same comment as above: repeating such string literals here makes the overall code brittle.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,531 @@
...
+
+  def extractHeap(h: Iterable[Chunk], model: Model): ExtractedHeap = {
+    var entries: List[HeapEntry] = List()
+    h foreach {
+      case c @ BasicChunk(FieldID, _, _, _, _) =>
+        val entry = extractField(c, model)
+        entries = entries :+ entry
+      case c @ BasicChunk(PredicateID, _, _, _, _) =>
+        val entry = extractPredicate(c, model)
+        entries = entries :+ entry
```
mschwerhoff, line 324: Probably doesn't matter much in practice in this case, but constructing a linked-list by repeatedly appending to its end is the worst-possible way w.r.t performance. A vector might be more suited here (see e.g. https://docs.scala-lang.org/overviews/collections-2.13/introduction.html).

src/main/scala/rules/SymbolicExecutionRules.scala
```diff
@@ -28,14 +28,18 @@ trait SymbolicExecutionRules {
...
-        val ce: Counterexample = if (Verifier.config.counterexample.toOption.get == "native") {
-          val oldHeap = if (s.oldHeaps.contains(Verifier.PRE_STATE_LABEL))
-            Some(s.oldHeaps(Verifier.PRE_STATE_LABEL).values)
-          else None
-          SiliconNativeCounterexample(s.g, s.h.values, oldHeap, nativeModel)
-        }else{
-          SiliconVariableCounterexample(s.g, nativeModel)
+        val ce_type = Verifier.config.counterexample()
+        val ce: Counterexample = ce_type match {
+          case "native" =>
```
mschwerhoff, line 33: Using an enum or a case object, instead of string literals, would facilitate maintenance

mschwerhoff (changes requested):
I've left several comments, but nothing severe.

Please make sure to properly benchmark the changes, at least against the Silver/Silicon test suite. Silicon [comes with a script](https://github.com/viperproject/silicon/blob/master/utils/scripts/benchmark.sh) that measures and processes runtime per example. You can use it to run old and new code (with counterexamples disabled); the resulting CSV files can then be analysed with Excel, Python, etc.

It might also be interesting for your thesis evaluation to compare the runtimes of (new) Silicon with and without counter examples, to see how much runtime overhead counter examples incur.

Discuss with your supervisor(s) first, though.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,531 @@
...
+              case (LitIntEntry(x), "-") => LitIntEntry(-x)
+              case _ => OtherEntry(s"$m", "not an integer literal")
+            }
+          case _ => OtherEntry(s"$m", "not an integer literal")
+        }
+      case sorts.Bool =>
+        m.toString.toLowerCase() match {
+          case "true"  => LitBoolEntry(true)
+          case "false" => LitBoolEntry(false)
+          case x =>
```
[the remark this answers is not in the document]
cedihegi: Well this was just me trying to play it safe, because I don't know if there are certain special cases and I would not want my code to fail because of that, ApplicationEntries of some sort like for the integers or I don't know. But if you say so I will certainly remove that part.

src/main/scala/rules/SymbolicExecutionRules.scala
cedihegi: Here I am a bit confused. This is the user input, i.e. the provided flag that comes with --counterexample <string> right? So at some point we would have to do this matching. Is your idea to do the string matching when constructing the config such that the config's field "ce_type" is an enum? Or am I completely misunderstanding your point?

mschwerhoff: Sorry, I should have been more explicit, but your guess is right: Scallop has a mechanism for converting "raw" options, e.g. arbitrary strings, into a more constrained form, e.g. values of an enum. This prevents unexpected option values from "leaking" beyond the option parser.

cedihegi: Okay but in this case this would have to happen on the silver level right since the counterexample type is parsed in SilFrontEndConfig.scala? Do you want me to open a new PR for that? I got it working but I am still not sure if this is what you expected, I assume this will require changes in Carbon too.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,506 @@
...
+  lazy val termconverter = {
+    val conv = new TermToSMTLib2Converter()
+    conv.start()
+    conv
+  }
+  //some tokens used for naming model entries in a more maintainable way
+  lazy val seqLengthId: String = {
+    val term = SeqLength(Var(Identifier("any"), sorts.Seq(sorts.Int)))
+    termconverter.convert(term).split(" ").head.substring(1)
+  }
```
cedihegi, line 136: @mschwerhoff so this is one example of how I used TermToSMTLib2Converter.convert() to get the String literals in a more maintainable way, but as you see from this example, it was not always possible for me to do it in a 100% maintainable way since I only use part of the returned String. So if just the names change, in this case Seq_length, this will still work but if the general structure changes there will still be a problem. Is this what you expected or do you think there is a better way to do this?

src/main/scala/rules/SymbolicExecutionRules.scala
mschwerhoff: Good point. In that case, leave the code as-is, but please add a TODO comment suggesting the change to an enum.

src/main/scala/reporting/Converter.scala
mschwerhoff: Thanks for the example! The code obscures your actual intention so much, that it is probably better to go with string literals for now — with them, it is at least obvious what the purpose of the code is.

A better way will probably require changes to Silicon's term converter, e.g. add public vals that hold the relevant sort and symbol names, which seems to go beyond the scope of your thesis.

So, just leave the code as-is, and add a TODO comment for how to improve the code.

cedihegi: Okay thanks, then I think I am pretty much done with the changes you suggested if you agree with them. 
But I struggle with the benchmarking for the silicon test suite. If I run your script on the silver tests it seems to produce something reasonable, but for silicon, even with the current master version, it states "The following output occurred during testing, but should not have according to the test annotations" or throws ParseErrors for almost every testcase. 
I tried both 
  ./utils/scripts/benchmark.sh src/test
  ./utils/scripts/benchmark.sh src/test/resources
the same commands with silvers test directory as target work fine. I made sure I have the most up to date version of both silver and silicon.

mschwerhoff: Not all tests in Silicon's `src/test/resources` are meant to be verified stand-alone, maybe that explains it. Due to the upcoming semester start I unfortunately can't help you with debugging, but if the Silver test suite runs fine and doesn't reveal any significant performance differences, that should be good enough.

Aurel300:
@mschwerhoff We weren't sure about your benchmarking suggestion – do you expect this change to make a difference to Silicon runtimes when counterexamples are not enabled? The `Converter` is only used when the new counterexample method is chosen, apart from that there is a single change of a `def` to `lazy val`.

Benchmarking the new counterexample method against the old one could work, though it's not particularly interesting data for the thesis itself. For the evaluation we are more interested in how useful the counterexample is in a Prusti setting, whether it accurately shows which variable is the culprit, etc.

Felalolf (commented):
I had one of my students try out some counter examples. He found one bug:
```
method test(xs: Seq[Int])
{
  assert |xs| > 3
  // option "mapped" outputs [0,0,0,0] as counterexample.
  // option "native" outputs a sequence of size 3. 
}
```

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,503 @@
...
+          "Seq_length",
+          Seq(ConstantEntry(name)),
+          sorts.Int
+        ).asInstanceOf[LitIntEntry].value.toInt
+      )
+      lenTry match {
+        case Success(x) =>
+          val len: Int = x
+          var values: Vector[ExtractedModelEntry] = Vector()
+          for (i <- 0 to len) {
```
Aurel300, line 362: This might be the cause of the bug @Felalolf commented.
```suggestion
          for (i <- 0 until len) {
```

cedihegi: Oh yes, thank you @Felalolf @Aurel300 ! I should have noticed that.

src/main/scala/reporting/Converter.scala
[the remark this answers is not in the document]
cedihegi: Okay so while using it for prusti I actually found an example where this triggers a match error now, without the third match case.
The problem in the model is:
`$SortWrappers.$SnapToBool -> {
    (= (:var 0) $Snap.unit)
}` 
Is this even something I could resolve to a boolean somehow?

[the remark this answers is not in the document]
marcoeilers: This means ``$SortWrappers.$SnapToBool(x) = (x == $Snap.unit)``.
For what it's worth, the ModelParser in Silver used to have special support for parsing this at least in some scenarios (see function boolFuncDef here: https://github.com/viperproject/silver/blob/4dbefcad60b98b80fbffd80e87a6355c79f07ae2/src/main/scala/viper/silver/verifier/ModelParser.scala) but this isn't in the current version any more for some reason.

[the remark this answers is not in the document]
cedihegi: Aah okay, thank you very much!
Then I guess I'll just re-implement this for my use if this is not used anywhere else.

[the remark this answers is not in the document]
marcoeilers: I'm not sure if that's what you meant, but IMO this should be implemented in the ModelParser in Silver again; the exisiting code in Nagini and 2Vyper for example also assume that it'ts there (and I think it was just removed by accident).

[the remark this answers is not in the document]
cedihegi: Oh no then it's not what I meant.
I guess I can do it but I've not worked on silver at all so far.

cedihegi:
@mschwerhoff I ran the benchmarks again and below there is a plot of the result.
But what exactly should I do with the data? 
And also, does this seem satisfying? The deviations are mostly small but there still seem to be some trends that I can not really explain.

plot-silicon

src/main/scala/reporting/Converter.scala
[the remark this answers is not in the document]
marcoeilers: I'll give it a try. Do you happen to have an example where Z3 generates this output so that I can use it for testing?

Aurel300:
@cedihegi Could you please mark as resolved the code review comments that you have addressed?

src/main/scala/reporting/Converter.scala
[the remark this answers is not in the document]
cedihegi: Yes sure.
[https://github.com/cedihegi/silicon/blob/counterexample-interface/src/test/resources/demo-counterexamples/prusti-bug.vpr](url)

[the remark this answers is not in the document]
marcoeilers: Thanks! Try this Silver branch: https://github.com/viperproject/silver/tree/marcoeilers_ce_parsing_improvement and let me know if something doesn't work.
I'll need to test it properly before I merge it into the main branch.

mschwerhoff:
> @mschwerhoff I ran the benchmarks again and below there is a plot of the result.
> But what exactly should I do with the data?
> And also, does this seem satisfying? The deviations are mostly small but there still seem to be some trends that I can not really explain.

I don't think I understand the plot. As before, I'd like to know that the new counterexample code, if the extraction feature is *disabled*, does not slow down Silicon. For that, it's typically useful to compare the runtime per example, and the overall runtime.

cedihegi:
> 
> 
> > @mschwerhoff I ran the benchmarks again and below there is a plot of the result.
> > But what exactly should I do with the data?
> > And also, does this seem satisfying? The deviations are mostly small but there still seem to be some trends that I can not really explain.
> 
> I don't think I understand the plot. As before, I'd like to know that the new counterexample code, if the extraction feature is _disabled_, does not slow down Silicon. For that, it's typically useful to compare the runtime per example, and the overall runtime.

I did run the benchmark on the counterexample code with counterexamples disabled and the current silicon branch as you said, and the plot shows the distribution for the increase or decrease of the runtimes  in miliseconds for each file. The points left of zero are the files where the counterexample branch was slower than the master branch, i.e. for each bin how many were slower by that amount of miliseconds. 
I was wondering if this plot indicates that there is a small slowdown in some cases and speedup in others, or if that is just run dependent and good enough and that's why I posted it. But I will run it again and find out.

marcoeilers:
Regarding performance, I just noticed that the code reading models from Z3 was horribly inefficient. I committed a fix (https://github.com/viperproject/silicon/commit/2641a668752108c0813e662f48af981d8bcefd24); if you have big counterexamples, that should improve performance and memory usage quite a bit.

Aurel300:
@mschwerhoff Ok, this should be ready to merge. If I understand correctly, @marcoeilers 's [branch](https://github.com/viperproject/silver/tree/marcoeilers_ce_parsing_improvement) should be merged first?

cedihegi:
One thing before merging, should I remove my testcases? There are not specifications and I guess they will cause problems. I kept them in so far for the people trying / reviewing my branch.

marcoeilers:
> @mschwerhoff Ok, this should be ready to merge. If I understand correctly, @marcoeilers 's [branch](https://github.com/viperproject/silver/tree/marcoeilers_ce_parsing_improvement) should be merged first?

My branch is merged in.
Regarding tests: IMO there should be something in the test suite that checks that this stuff doesn't break, e.g., we could turn on counterexample parsing by default for the normal tests. Is there something like that at the moment?

cedihegi:
> > @mschwerhoff Ok, this should be ready to merge. If I understand correctly, @marcoeilers 's [branch](https://github.com/viperproject/silver/tree/marcoeilers_ce_parsing_improvement) should be merged first?
> 
> My branch is merged in.
> Regarding tests: IMO there should be something in the test suite that checks that this stuff doesn't break, e.g., we could turn on counterexample parsing by default for the normal tests. Is there something like that at the moment?

@marcoeilers I am sorry for not answering for so long if this question was meant for me, but I don't know of anything like that, I also don't know if there is anything similar for the other counterexample classes.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,503 @@
...
+import viper.silicon.interfaces.state.Chunk
+import viper.silicon.resources.{FieldID, PredicateID}
+import viper.silicon.state.{Store, State, BasicChunk}
+import viper.silicon.state.terms._
+import viper.silicon.decider.TermToSMTLib2Converter
+
+//Classes for extracted Model Entries
+case class ExtractedModel(entries: Map[String, ExtractedModelEntry]) {
+  override lazy val toString: String = {
+    entries.map(x => s"${x._1} <- ${x._2.toString}").mkString("\n")
```
mschwerhoff, line 15: Iteration order of `Map` is probably nondeterministic, resulting in a nondeterministic `toString` (across multiple runs), which may be confusing. You could use a deterministic alternative map (see https://github.com/viperproject/silicon/blob/master/src/main/scala/package.scala#L17), or sort by keys before creating the string.

src/main/scala/reporting/Converter.scala
```diff
@@ -0,0 +1,503 @@
...
+      ConstantEntry(value.toString)
+    }
+  }
+}
+
+case class RefEntry(
+    name: String,
+    fields: Map[String, (ExtractedModelEntry, Option[Rational])]
+) extends ExtractedModelEntry {
+  override lazy val toString: String = {
```
mschwerhoff, line 55: Potentially non-det

mschwerhoff (approved):
Looks good, but watch out for nondeterministic strings.

src/main/scala/reporting/Converter.scala
cedihegi: https://github.com/viperproject/silicon/blob/0c495c0fc14f2b22f6710167ab4b86a076f46790/src/main/scala/package.scala#L17-L18

I think the keyword `Map` here refers to the above type definition. In this case the order should be deterministic, right? At least that was my understanding when I coded this, because the implementation using this in Prusti also relies on these maps being deterministic and in my use-cases they were.

mschwerhoff: IntelliJ will tell you, if you hover over the type or go to its definition, if I'm not mistaken. If so, you can indeed ignore my comment.

cedihegi: Yeah I was wrong, the definition was taken from `scala.Predef`. But it is fixed now.

mschwerhoff (approved):

PR #542, opened 2021-04-17
Add field reporter to trait Verifier - adapting Silicon

aterga:
Should be merged after [Silver PR 513](https://github.com/viperproject/silver/pull/513). 

Note that here we want ```Reporter``` as opposed to ```PluginAwareReporter``` (which Silicon used before) as the latter is needed only for Silver plugins; I think that the verifiers don't have to know about the Silver plugin system.


mschwerhoff (approved):
Looks all good to me

PR #546, opened 2021-04-24
WIP: Report one Error per Path

tschoesi:
This is a draft PR meant to get some feedback from @mschwerhoff (of course, others are welcome to comment as well!)

This PR contains a wrapper for the verificationresults, so that instead of short circuiting when silicon branches, the results from both branches can be reported.

Todo/Questions:

- [ ] Some of the original tests don't pass anymore because it now reports more errors, should I also adapt those?
- [ ] Add a couple more tests

### DIFF
--- a/src/main/scala/interfaces/Verification.scala
+++ b/src/main/scala/interfaces/Verification.scala
@@ -7,13 +7,52 @@
 package viper.silicon.interfaces
 
 import viper.silicon.interfaces.state.Chunk
+import viper.silicon.state.terms.Term
 import viper.silicon.state.{State, Store}
 import viper.silver.verifier.{Counterexample, Model, VerificationError}
-import viper.silicon.state.terms.Term
 
 /*
  * Results
  */
+class VerificationResultWrapper(verificationResultsarg: Seq[VerificationResult]) {
+  def verificationResults: Seq[VerificationResult] = verificationResultsarg
+
+  def &&(other: => VerificationResultWrapper): VerificationResultWrapper = {
+    //Filter out multiple Successes
+    var combinedSeq = this.verificationResults ++ other.verificationResults
+    val hasSuccess = combinedSeq.contains(Success())
+    combinedSeq = combinedSeq flatMap {
+      case _: Success => None
+      case vr: VerificationResult => Some(vr)
+    }
+    if (combinedSeq.isEmpty && hasSuccess) combinedSeq = Seq(Success())
+    new VerificationResultWrapper(combinedSeq)
+  }
+  // short circuit "and"
+  def &&&(other: => VerificationResultWrapper): VerificationResultWrapper = { //TODO:J check if this is really needed
+    if (this.containsFatal) this else this && other
+  }
+  def getFatals: Seq[FatalResult] = verificationResults flatMap{
+     case fr: FatalResult => Some(fr)
+     case _ => None
+  }
+  def getFailures: Seq[Failure] = verificationResults flatMap {
+    case fl: Failure => Some(fl)
+    case _ => None
+  }
+  def containsFatal: Boolean = verificationResults.foldLeft(false)((b: Boolean,v: VerificationResult)=>b || v.isFatal)
+  def containsFailure: Boolean = getFailures.nonEmpty
+}
+
+object VerificationResultWrapper {
+  def apply(vr: VerificationResult): VerificationResultWrapper = new VerificationResultWrapper(Seq(vr))
+  def apply(vrs: Seq[VerificationResult]): VerificationResultWrapper = new VerificationResultWrapper(vrs)
+
+  def unapply(arg: VerificationResultWrapper): Option[Seq[VerificationResult]] = Some(arg.verificationResults)
+//  def unapply(vrw: VerificationResultWrapper): Option[Seq[VerificationResult]] = if (vrw.verificationResults.nonEmpty) Some(vrw.verificationResults) else None
+//  def unapply(vr: VerificationResult): Option[VerificationResult] = Some(vr) //TODO:J check how this could work
+}
+
 
 /* TODO: Extract appropriate interfaces and then move the implementations
  *       outside of the interfaces package.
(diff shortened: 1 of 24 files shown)


mschwerhoff:
>     * [ ]  Some of the original tests don't pass anymore because it now reports more errors, should I also adapt those?

Yes: please adapt them, create a WIP PR for Silver as well, and let the two dependent PRs refer to each other via links in the PR description.

src/main/scala/interfaces/Verification.scala
```diff
@@ -7,13 +7,52 @@
...
 import viper.silicon.interfaces.state.Chunk
+import viper.silicon.state.terms.Term
 import viper.silicon.state.{State, Store}
 import viper.silver.verifier.{Counterexample, Model, VerificationError}
-import viper.silicon.state.terms.Term
 
 /*
  * Results
  */
+class VerificationResultWrapper(verificationResultsarg: Seq[VerificationResult]) {
```
mschwerhoff, line 17: Please replace `verificationResultsarg` and accompanying `def verificationResults` by just parameter `val verificationResults`

src/main/scala/interfaces/Verification.scala
```diff
@@ -7,13 +7,52 @@
...
+    var combinedSeq = this.verificationResults ++ other.verificationResults
+    val hasSuccess = combinedSeq.contains(Success())
+    combinedSeq = combinedSeq flatMap {
+      case _: Success => None
+      case vr: VerificationResult => Some(vr)
+    }
+    if (combinedSeq.isEmpty && hasSuccess) combinedSeq = Seq(Success())
+    new VerificationResultWrapper(combinedSeq)
+  }
+  // short circuit "and"
```
mschwerhoff, line 31: Leave blank lines between methods

src/main/scala/interfaces/Verification.scala
```diff
@@ -7,13 +7,52 @@
...
+  def &&&(other: => VerificationResultWrapper): VerificationResultWrapper = { //TODO:J check if this is really needed
```
mschwerhoff, line 32: While method/operator `&&` conveys some intuition, it will be difficult to make an educated guess about the difference between `&&` and `&&&`. Is `&&&` really necessary as an operator? Can it be given a more self-explaining name?

src/main/scala/interfaces/Verification.scala
```diff
@@ -7,13 +7,52 @@
...
+  }
+  def getFatals: Seq[FatalResult] = verificationResults flatMap{
+     case fr: FatalResult => Some(fr)
+     case _ => None
+  }
+  def getFailures: Seq[Failure] = verificationResults flatMap {
+    case fl: Failure => Some(fl)
+    case _ => None
+  }
+  def containsFatal: Boolean = verificationResults.foldLeft(false)((b: Boolean,v: VerificationResult)=>b || v.isFatal)
```
mschwerhoff, line 43: Try `verificationResults.exists`

src/main/scala/interfaces/Verification.scala
```diff
@@ -7,13 +7,52 @@
...
+  def containsFailure: Boolean = getFailures.nonEmpty
+}
+
+object VerificationResultWrapper {
+  def apply(vr: VerificationResult): VerificationResultWrapper = new VerificationResultWrapper(Seq(vr))
+  def apply(vrs: Seq[VerificationResult]): VerificationResultWrapper = new VerificationResultWrapper(vrs)
+
+  def unapply(arg: VerificationResultWrapper): Option[Seq[VerificationResult]] = Some(arg.verificationResults)
```
mschwerhoff, line 51: `arg` is really not the world's greatest variable name

src/main/scala/interfaces/Verification.scala
mschwerhoff, line 17: A few things bug me about this class, and I have the impression that a better design is possible when taking immutability into account. `&&`, `getFatals`, `getFailures` and `containsFatal` all filter/partition the results, which seems potentially redundant if many calls are made.

Since parameter `verificationResultsarg` is an immutable sequence of immutable results, it would suffice to partition the incoming results into failures and successes, and to store the two separately. This should reduce the number of subsequent filter steps necessary in aforementioned methods.

Moreover, all methods but `&&` could be turned into `lazy val`s instead of `def`s, since subsequent calls should always return the same results anyway.

src/main/scala/rules/Brancher.scala
```diff
@@ -1,190 +1,189 @@
...
-            throw ex
-        }
-
-        assert(rs.length == 1, s"Expected a single verification result but found ${rs.length}")
-        rs.head
-      }
-    }
-  }
-}
+// This Source Code Form is subject to the terms of the Mozilla Public
```
mschwerhoff, line 1: Try to minimise the diffs, and e.g. revert any whitespace changes. Did anything in `Brancher.scala` actually change?

src/main/scala/rules/ChunkSupporter.scala
```diff
@@ -26,12 +26,12 @@ trait ChunkSupportRules extends SymbolicExecutionRules {
               ve: VerificationError,
               v: Verifier,
               description: String)
-             (Q: (State, Heap, Term, Verifier) => VerificationResult)
-             : VerificationResult
+             (Q: (State, Heap, Term, Verifier) => VerificationResultWrapper)
```
mschwerhoff, line 29: Changing from `VerificationResult` to `VerificationResultWrapper` affects loads of public APIs and may force clients/frontends to perform similar boilerplate changes.

Please think about changing `VerificationResult` itself — would this be a reasonable alternative?

mschwerhoff (commented):
A first batch of comments. Please let me know if you'd like to discuss them, or when you've addressed them.

src/main/scala/rules/Brancher.scala
tschoesi: This might be because I use `LF` as the line separator and the file was using `CRLF` before, I'll change it so hopefully it doesn't show the whole file as changed!

PR #549, opened 2021-05-02
WIP: Report one error per path

tschoesi:
Another try at getting silicon to report one error per path, this time with fewer boilerplate code changes.
Again, this is meant to get feedback from @mschwerhoff .

Two tests still fail (report errors that, according to the annotations, should not be reported)
* [list_insert_tmp.vpr](https://github.com/viperproject/silver/blob/b2f9190a9794bcc31ce93f580c1aff2dbfcc2a07/src/test/resources/wands/examples/list_insert_tmp.vpr#L132)
* [QPWands.vpr](https://github.com/viperproject/silver/blob/83ff132b847fd97ef20e1b622167de706d52a29a/src/test/resources/wands/new_syntax/QPWands.vpr#L38)

I'm currently reading up on magic wands, as those are both magic wand related tests (right now I'm not yet sure if those errors should be reported or not..)

### DIFF
--- a/src/main/scala/interfaces/Verification.scala
+++ b/src/main/scala/interfaces/Verification.scala
@@ -21,25 +21,22 @@ import viper.silicon.state.terms.Term
 
 /* TODO: Make VerificationResult immutable */
 sealed abstract class VerificationResult {
-  var previous: Option[NonFatalResult] = None
+  var previous: Set[VerificationResult] = Set()
 
   def isFatal: Boolean
   def &&(other: => VerificationResult): VerificationResult
 
-  def allPrevious: List[VerificationResult] =
-    previous match {
-      case None => Nil
-      case Some(vr) => vr :: vr.allPrevious
-    }
-
-  def append(other: NonFatalResult): VerificationResult =
-    previous match {
-      case None =>
-        this.previous = Some(other)
+  def combine(other: => VerificationResult): VerificationResult = {
+    val r: VerificationResult = other
+    this match {
+      case _ : FatalResult =>
+        this.previous = this.previous + r ++  r.previous
         this
-      case Some(vr) =>
-        vr.append(other)
+      case _ =>
+        r.previous = r.previous + this ++ this.previous
+        r
     }
+  }
 }
 
 sealed abstract class FatalResult extends VerificationResult {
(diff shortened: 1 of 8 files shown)


src/main/scala/interfaces/Verification.scala
```diff
@@ -21,25 +21,22 @@ import viper.silicon.state.terms.Term
 
 /* TODO: Make VerificationResult immutable */
 sealed abstract class VerificationResult {
-  var previous: Option[NonFatalResult] = None
+  var previous: Set[VerificationResult] = Set()
```
mschwerhoff, line 24: Keep in mind that sets typically have non-deterministic iteration order, which may be undesirable (e.g. maybe we want earlier results to also occur earlier in the collection?) or confusing (e.g. when outputting the results without sorting them first).

If necessary, alternatives would be set instance that preserves insertion order or are sorted/ordered, or using another data structure, e.g. a vector.

src/main/scala/interfaces/Verification.scala
```diff
@@ -21,25 +21,22 @@ import viper.silicon.state.terms.Term
...
-      case None => Nil
-      case Some(vr) => vr :: vr.allPrevious
-    }
-
-  def append(other: NonFatalResult): VerificationResult =
-    previous match {
-      case None =>
-        this.previous = Some(other)
+  def combine(other: => VerificationResult): VerificationResult = {
+    val r: VerificationResult = other
```
mschwerhoff, line 30: Please add a comment that explains that the line here executes `other`. This is subtle and often overlooked, in particular by people not regularly working with Scala.

mschwerhoff (commented):
Looks good, modulo minor comments

src/main/scala/interfaces/Verification.scala
tschoesi: Added this now.

src/main/scala/interfaces/Verification.scala
tschoesi: As discussed, this will be changed if/when it will be needed.

src/main/scala/interfaces/Verification.scala
tschoesi: Added this now

mschwerhoff: Was that committed? I can't see it here.

tschoesi: I think it somehow still shows the old diff here, but it's in [this commit](https://github.com/tschoesi/silicon/commit/a19df8b0c6684de33cd35eda37429c705c82f6be).

PR #555, opened 2021-06-14
Declaring SortWrapper functions for backend types

marcoeilers:
This PR fixes errors that occur when Silicon uses SortWrapper functions from and to backend types (float and bitvector); previously, these functions were not declared, and using the functions resulted in Z3 errors (Vytautas had an example where this happened; there is a corresponding branch in Silver that adds a test case for this). 

It adds code that checks which backend types are used, and declares the corresponding SortWrapper functions. The change in ``sortwrappers.smt2`` is because for backend type names, which can for example contain spaces (e.g. ``(_ FloatingPoint  8  24)``), we need to differentiate between the name of the type name itself (``$T$``) and a sanitized version of the type that does not contain spaces (``$S$``), which is used in the name of the SortWrapper function.

### DIFF
--- a/src/main/scala/verifier/DefaultMasterVerifier.scala
+++ b/src/main/scala/verifier/DefaultMasterVerifier.scala
@@ -25,9 +25,12 @@ import viper.silicon.supporters._
 import viper.silicon.supporters.functions.DefaultFunctionVerificationUnitProvider
 import viper.silicon.supporters.qps._
 import viper.silicon.utils.Counter
+import viper.silver.ast.{BackendFunc, BackendType}
 import viper.silver.ast.utility.rewriter.Traverse
 import viper.silver.cfg.silver.SilverCfg
 import viper.silver.reporter.{ConfigurationConfirmation, Reporter, VerificationResultMessage}
+
+import scala.collection.mutable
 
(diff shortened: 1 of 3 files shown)


src/main/scala/verifier/DefaultMasterVerifier.scala
```diff
@@ -395,6 +398,12 @@ class DefaultMasterVerifier(config: Config, override val reporter: Reporter)
     sortWrapperDeclarationOrder foreach (component =>
       emitSortWrappers(component.sortsAfterAnalysis, sink))
 
+    val backendTypes = new mutable.HashSet[BackendType]
```
mschwerhoff, line 401: Is that a deterministic collection? If not, we might be better of with a `LinkedHashset`, see https://github.com/viperproject/silicon/blob/master/src/main/scala/package.scala#L31.

mschwerhoff (approved):
Approved, but see my comment about deterministic hash sets.

src/main/scala/verifier/DefaultMasterVerifier.scala
marcoeilers: Done

PR #475, opened 2020-02-20
Scope-based SymbExLogger

viper-admin:
>  **Pull request** :twisted_rightwards_arrows: created by **@arquintl** on 2019-08-23 11:45
> Last updated on 2020-02-14 14:40
> Original Bitbucket pull request id: 65
>
> Participants:
>
> * **@mschwerhoff** (reviewer)
> * **@arquintl**
>
> Source: https://github.com/viperproject/silicon/commit/5670297dad2fff27c5dd36ba25b413839d0bf3c5 on branch `arquintl/silicon/default`
> Destination: https://github.com/viperproject/silicon/commit/76f21ec30b0fe9e7be00e4e6f56db9fd42984a42 on branch `master`
>
> State: **`OPEN`**

* adds `logConfig` as additional Silicon config parameter
* expects `consolidateOnAssertTrue` as additional SilFrontendConfig parameter \(TODO LA: PR for silver\)
* adds implementation of new SymbExLogger with heavy code cleanup \(it’s now more than 1 source file → viper.silicon.logger package
* moves unit tests of SymbExLogger to a dedicated test file: SymbExLoggerTests
* makes SymbExLogger unit tests more flexible by ignoring additional records and expecting only method, function, predicate, braching, joining, and execute records

PR #562, opened 2021-07-06
Scope-based SymbExLogger Take 2

ArquintL:
Introduces the same changes as #475.
Apparently, the changes introduced by the other PR have been added to master in commit 85ecb641a28f0a58808528d520aaa088a2b8dbf8 but apparently undone by accident in the next commit 73f7fd162e6876d595ab88035527b2fb7b995385.

This PR is almost identical except:
- `writeLogFile` seems to no longer exist in the Silver codebase and thus the SymbExLogger will only use `ideModeAdvanced` (instead of looking at both flags) to decide whether it should be enabled or not
- file validation has been enabled for the file/path passed as `logConfig`
- some expected outputs for SymbExLogger tests had to be adapted. Looking two years back, it proved to be relatively robust:
  - Silicon does not seem to perform a Join operation in some cases after branching anymore, thus the corresponding Join record had to be removed
  - Caused by the change above, the numbering of some Silicon internal variables changes in the log output (they are used as path conditions). However, this could be made more robust in the future by e.g. extracting `$t` from `$t@6@01`
  - Silicon seems to print fields slightly different. This shows up in the `symbLogTest_linkedListMinimal`: previously, a branch has been logged as `Branch $t@8@01 != Null:` and nowadays it is `Branch First:($t@6@01) != Null:`

@aterga: these are the changes performed as part of my Master's thesis. I cannot recall whether we still support the old logging format and/or whether/how clients would need to be adapted. Could you maybe give it a try?


ArquintL:
All Silicon tests pass locally

mschwerhoff (approved):
Looks good, thank you.

PR #566, opened 2021-07-14
Flyweight ASTs

fabianboesiger:
Adds macro code to create flyweight terms, which are not used in the master branch as it did not increase performance as expected. However the macro implementation may be used for different purposes in the future.

### DIFF
--- a/annotation/README.md
+++ b/annotation/README.md
@@ -5,3 +5,15 @@ This directory contains the macro code to generate flyweight terms.
 Flyweight ASTs are not used in the master branch as it did not increase performance as expected and thus is of no practical use. However the macro implementation may be used for different purposes in the future.
 
 The file `Terms.example.scala` in this directory shows how the flyweight annotation can be used in practice.
+
+If ever needed again, the `annotation` subproject  needs to be included in Silicon's `build.sbt` again:
+
+```
+lazy val annotation = project in file("annotation")
+// ...
+
+// Silicon specific project settings
+lazy val silicon = (project in file("."))
+  .dependsOn(annotation)
+  // ...
+```
--- a/annotation/README.md
+++ b/annotation/README.md
@@ -1,0 +1,7 @@
+# Flyweight ASTs
+
+This directory contains the macro code to generate flyweight terms.
+
+Flyweight ASTs are not used in the master branch as it did not increase performance as expected and thus is of no practical use. However the macro implementation may be used for different purposes in the future.
+
+The file `Terms.example.scala` in this directory shows how the flyweight annotation can be used in practice.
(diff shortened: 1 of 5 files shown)


annotation/README.md
```diff
@@ -0,0 +1,19 @@
+# Flyweight ASTs
+
+This directory contains the macro code to generate flyweight terms.
+
+Flyweight ASTs are not used in the master branch as it did not increase performance as expected and thus is of no practical use. However the macro implementation may be used for different purposes in the future.
```
mschwerhoff, line 5: Please add a reference to your thesis (e.g. your name and the thesis title), to provide some contextual information.

mschwerhoff (approved):
Looks good, thank you. Please double-check that all files are UTF-8 encoded (no BOM) and use Linux file endings. If that's the case, I'll merge the PR.

annotation/README.md
fabianboesiger: Regarding the thesis name, I think a new name would be appropriate as "Flyweight ASTs" is only relevant to the first part.

My idea was something like "Switching up Silicon" as the thesis title, with part 1 "Flyweight ASTs; A Study in Applied Laziness", and part 2 "Joining; Getting Rid of the Branches".

I assume all the files are UTF-8 encoded and use Linux file endings as I used Linux myself, but is there a quick way to check multiple files?

fabianboesiger: I checked all files using the `file` utility. I couldn't find any CRLF line endings, and all files show up as UTF-8 or ASCII, so i assume this is all good.

mschwerhoff: For now, just add something like "... BSc thesis of Fabian Bösiger ..." to the text, so that it's documented where the evaluation results (no improvement) come from.

Regarding line endings, see https://github.com/viperproject/silicon/blob/master/utils/scripts/lineendings.sh. Not sure about UTF-8, but I'm sure you'll quickly find suitable shell magic on Stackoverflow etc.

Regarding your thesis title: if a potential employer reads "Switching up Silicon", they most likely will get no intuition for your thesis (switching? Silicon?). I'd therefore go for something more descriptive, even if less catchy. "Improving performance of a program verifier from two different angles" is definitely too long, but illustrates what I mean by "descriptive".

Aurel300: `for f in file1 file2 file3; do echo "$f"; xxd "$f" | head -n 1; done`

Then see if anything begins with 2 or 3 high non-ASCII bytes.

`grep -U $'\x0D' file1 file2 file3`

No output would indicate no CR bytes.

fabianboesiger: @mschwerhoff We can discuss a thesis title in the next meeting, until we know more I will refer to the thesis by my name.

fabianboesiger: @Aurel300 thanks for the quick help. No CR bytes show up, however I am a bit confused by your first suggested command. How do I interpret such an output:
```
annotation.iml
00000000: 3c3f 786d 6c20 7665 7273 696f 6e3d 2231  <?xml version="1
```

Aurel300: Yup, that's fine, there is no BOM. The command shows the first 16 bytes of the file. If there was a UTF-8 BOM, the first 3 bytes would be `efbbbf`.

fabianboesiger: Ah thanks for your help. All files are good, then!

PR #568, opened 2021-07-16
Change double quotes to curly braces as delimeters for composite command line options

aterga:
This PR changes the expected format of arguments passed to ```--z3ConfigArgs```. 

We used to use double quotes, but this caused problems with escaping the ```"``` characters in case the silicon command itself had to be passed as a string. In particular, the only way I found to use this option on macOS in bash and zsh was like this: 

```
java -cp silicon.jar viper.silicon.SiliconRunner --z3ConfigArgs "\"model.partial=true\"" example.vpr
```
However, I still couldn't figure out how to encode the escapements themselves s.t. they could be added to the Viper IDE configuration. It seems that avoiding double quotes in the first place would simplify things quite a bit.

Therefore, I propose the following syntax: 
```
--z3ConfigArgs {arg1=val1 arg2=val2 ...}
``` 

I tested this on my local machine with macOS; I would appreciate it if you could try it out on Windows and Linux as well before merging this PR (not urgent).

### DIFF
--- a/src/main/scala/Config.scala
+++ b/src/main/scala/Config.scala
@@ -54,7 +54,7 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
 
   private val smtlibOptionsConverter: ValueConverter[Map[String, String]] = new ValueConverter[Map[String, String]] {
     def parse(s: List[(String, List[String])]): Either[String, Option[Map[String, String]]] = s match {
-      case (_, str :: Nil) :: Nil if str.head == '"' && str.last == '"' =>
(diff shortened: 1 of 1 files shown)


mschwerhoff:
I'm not too keen on changing the syntax, unless it's clear that escaping is impossible (or really gets out of hand). Maybe the syntax shown here helps: https://github.com/viperproject/silicon/wiki/Running-Silicon.

Since you already implemented a suggestion: does it work with passing options through sbt (-Dsilicon:key=value)?

mschwerhoff:
@aterga Any news?

aterga:
It seems that @fpoli 's solution solves the issue that this PR was trying to solve, too. Feel free to close if you agree

PR #572, opened 2021-07-26
More Joins

fabianboesiger:
Adds the config option `--moreJoins` which enables joining on impure branches for implications, conditional expressions, and also on branches caused by if-statements.


src/main/scala/rules/Consumer.scala
```diff
@@ -199,6 +236,44 @@ object consumer extends ConsumptionRules {
...
+            val s2 = entries match {
+              case Seq(entry) => // One branch is dead
+                (entry.s, entry.data)
+              case Seq(entry1, entry2) => // Both branches are alive
+                val mergedData = (
+                  State.mergeHeap(
+                    entry1.data._1, And(entry1.pathConditions.branchConditions),
+                    entry2.data._1, And(entry2.pathConditions.branchConditions),
+                  ),
+                  // Asume that entry1.pcs is inverse of entry2.pcs
```
mschwerhoff, line 264: Unclear: do you create an SMT assumption here, or is there some assumption underlying the next line of code? It is also not clear how sets of path conditions can be inverses of each other.

Typo in "asume"

src/main/scala/rules/Consumer.scala
```diff
@@ -199,6 +236,44 @@ object consumer extends ConsumptionRules {
...
+              (s2, v2) => consumeR(s2, h, a2, pve, v2)((s3, h1, t1, v3) => {
+                SymbExLogger.currentLog().closeScope(uidCondExp)
+                QB(s3, (h1, t1), v3)
+              }))
+          })(entries => {
+            val s2 = entries match {
+              case Seq(entry) => // One branch is dead
+                (entry.s, entry.data)
+              case Seq(entry1, entry2) => // Both branches are alive
+                val mergedData = (
```
mschwerhoff, line 259: Why is `State.mergeHeap` necessary here, whereas just `entry1.pathConditionAwareMerge` suffices in other places, e.g. for performing a CFG join?

src/main/scala/rules/Executor.scala
```diff
@@ -46,7 +47,7 @@ object executor extends ExecutionRules {
   import evaluator._
   import producer._
 
-  private def follow(s: State, edge: SilverEdge, v: Verifier)
+  private def follow(s: State, edge: SilverEdge, v: Verifier, joinPoint: Option[SilverBlock])
```
mschwerhoff, line 50: Maybe worthwhile: if you define the last argument to be `joinPoint: Option[SilverBlock] = None`, then no argument has to be explicitly provided at call site in the `None` case.

src/main/scala/rules/Executor.scala
```diff
@@ -63,74 +64,136 @@ object executor extends ExecutionRules {
...
+      }
     }
   }
 
   private def follows(s: State,
                       edges: Seq[SilverEdge],
                       @unused pvef: ast.Exp => PartialVerificationError,
-                      v: Verifier)
+                      v: Verifier,
+                      joinPoint: Option[SilverBlock])
```
mschwerhoff, line 104: See comment about `... = None` above (if worthwhile doing)

src/main/scala/rules/Executor.scala
```diff
@@ -63,74 +64,136 @@ object executor extends ExecutionRules {
...
-          SymbExLogger.currentLog().markReachable(uidBranchPoint)
-          follow(s, edge, v)(Q)
+          SymbExLogger.currentLog().endBranchPoint(uidBranchPoint)
+          res
       }
-      SymbExLogger.currentLog().endBranchPoint(uidBranchPoint)
-      res
+
+    } else {
+      sys.error("At most two out edges expected.")
```
mschwerhoff, line 178: Without joining, Silicon can handle more than two outgoing edges. Can you reasonably restructure the code so that the error is only raised when Option `moreJoins` is enabled?

src/main/scala/rules/Executor.scala
```diff
@@ -63,74 +64,136 @@ object executor extends ExecutionRules {
...
+      assert(edge1.source == edge2.source)
+      val branchPoint = edge1.source
+
+      s.methodCfg.joinPoints.get(branchPoint) match {
+        // If a join point was found, the config argument is enabled,
+        // and the source of both conditional edges was a statement block
+        // (we are at an if-statement), we may join the resulting branches again.
+        case Some(newJoinPoint) if
+          Verifier.config.moreJoins() &&
+          edge1.isInstanceOf[ConditionalEdge[ast.Stmt, ast.Exp]] &&
```
mschwerhoff, line 125: Aren't the `isInstanceOf` checks redundant, due to the casts on lines 113 and 114?

src/main/scala/rules/Executor.scala
```diff
@@ -63,74 +64,136 @@ object executor extends ExecutionRules {
...
+          edge2.isInstanceOf[ConditionalEdge[ast.Stmt, ast.Exp]] &&
+          edge1.source.isInstanceOf[StatementBlock[ast.Stmt, ast.Exp]] &&
```
mschwerhoff, line 127: Why are the restrictions on lines 127 and 128 necessary? Please explain in code comments.

src/main/scala/rules/Executor.scala
```diff
@@ -63,74 +64,136 @@ object executor extends ExecutionRules {
...
+            case (exp1, ast.Not(exp2)) => exp1 == exp2
+            case (ast.Not(exp1), exp2) => exp1 == exp2
+            case _ => false
+          })
+
+          eval(s, edge1.condition, pvef(edge1.condition), v)((s1, t0, v1) =>
+            joiner.join[scala.Null, scala.Null](s1, v1, resetState = false)((s2, v2, QB) => {
+              brancher.branch(s2, t0, v2)(
+                // Follow only until join point.
+                (s3, v3) => follow(s3, edge1, v3, Some(newJoinPoint))((s, v) => QB(s, null, v)),
```
mschwerhoff, line 141: Add a code comment stating why passing `null` is fine here

src/main/scala/rules/Executor.scala
```diff
@@ -63,74 +64,136 @@ object executor extends ExecutionRules {
...
   def exec(s: State, graph: SilverCfg, v: Verifier)
           (Q: (State, Verifier) => VerificationResult)
           : VerificationResult = {
 
-    exec(s, graph.entry, cfg.Kind.Normal, v)(Q)
+    exec(s, graph.entry, cfg.Kind.Normal, v, None)(Q)
   }
 
-  def exec(s: State, block: SilverBlock, incomingEdgeKind: cfg.Kind.Value, v: Verifier)
+  def exec(s: State, block: SilverBlock, incomingEdgeKind: cfg.Kind.Value, v: Verifier, joinPoint: Option[SilverBlock])
```
mschwerhoff, line 189: See earlier `... = None` comments

src/main/scala/rules/Joiner.scala
```diff
@@ -13,18 +13,24 @@ import viper.silicon.logger.records.structural.JoiningRecord
 import viper.silicon.state.State
 import viper.silicon.verifier.Verifier
 
-case class JoinDataEntry[D](s: State, data: D, pathConditions: RecordedPathConditions)
+case class JoinDataEntry[D](s: State, data: D, pathConditions: RecordedPathConditions) {
+  // Instead of merging states, we can directly merge JoinDataEntries to obtain new States.
```
mschwerhoff, line 17: Is this a "future work" comment? Otherwise, I'm confused: you say "instead of merging states", but then have `State.merge` on line 20.

src/main/scala/rules/Joiner.scala
```diff
@@ -40,15 +46,25 @@ object joiner extends JoiningRules {
...
+            s3.copy(g = s1.g,
+                    h = s1.h,
+                    oldHeaps = s1.oldHeaps,
+                    underJoin = s1.underJoin,
+                    // TODO: Evaluation should not affect partiallyConsumedHeap, probably
+                    ssCache = s1.ssCache,
+                    partiallyConsumedHeap = s1.partiallyConsumedHeap,
+                    invariantContexts = s1.invariantContexts)
+          } else {
+            // For more joins, state shouldn't be reset.
```
mschwerhoff, line 65: Does this express the assumption that `resetState == Config.moreJoins()`? If so, why is `resetState` needed?

src/main/scala/state/State.scala
```diff
@@ -198,6 +200,214 @@ object State {
...
+    sys.error(s"State merging failed: unexpected mismatch between symbolic states: $err")
+  }
+
+  // Merges two maps based on fOnce, if entry only exists in one map,
+  // and fTwice if entry exists in both maps.
+  // Used to merge the Store.
+  private def mergeMaps[K, V, D](map1: Map[K, V], data1: D, map2: Map[K, V], data2: D)
+                                (fOnce: (V, D) => Option[V])
+                                (fTwice: (V, D, V, D) => Option[V])
+  : Map[K, V] = {
```
mschwerhoff, line 225: Please align `:` with opening parenthesis from argument list(s) (as done elsewhere)

src/main/scala/state/State.scala
```diff
@@ -198,6 +200,214 @@ object State {
...
+  def mergeHeap(h1: Heap, cond1: Term, h2: Heap, cond2: Term): Heap = {
+    val (unconditionalHeapChunks, h1HeapChunksToConditionalize) = h1.values.partition(c1 => h2.values.exists(_ == c1))
+    val h2HeapChunksToConditionalize = h2.values.filter(c2 => !unconditionalHeapChunks.exists(_ == c2))
+    val h1ConditionalizedHeapChunks = conditionalizeChunks(h1HeapChunksToConditionalize, cond1)
+    val h2ConditionalizedHeapChunks = conditionalizeChunks(h2HeapChunksToConditionalize, cond2)
+    Heap(unconditionalHeapChunks) + Heap(h1ConditionalizedHeapChunks) + Heap(h2ConditionalizedHeapChunks)
+  }
+
+  def merge(s1: State, pc1: RecordedPathConditions, s2: State, pc2: RecordedPathConditions): State = {
+    //println("Merge")
```
mschwerhoff, line 268: Remove comment

src/main/scala/state/State.scala
```diff
@@ -198,6 +200,214 @@ object State {
...
+    s1 match {
+      /* Decompose state s1 */
+      case State(g1, h1, oldHeaps1,
+      parallelizeBranches1,
```
mschwerhoff, line 272: Indentation

src/main/scala/state/State.scala
```diff
@@ -198,6 +200,214 @@ object State {
...
+      reserveHeaps1, reserveCfgs1, conservedPcs1, recordPcs1, exhaleExt1,
+      applyHeuristics1, heuristicsDepth1, triggerAction1,
+      ssCache1, hackIssue387DisablePermissionConsumption1,
+      qpFields1, qpPredicates1, qpMagicWands1, smCache1, pmCache1, smDomainNeeded1,
+      predicateSnapMap1, predicateFormalVarMap1, hack) =>
+
+        /* Decompose state s2: most values must match those of s1 */
+        s2 match {
+          case State(g2, h2, oldHeaps2,
+          `parallelizeBranches1`,
```
mschwerhoff, line 294: Indentation

src/main/scala/state/State.scala
```diff
@@ -198,6 +200,214 @@ object State {
...
+              .map({case (h1, h2) => mergeHeap(h1, conditions1, h2, conditions2)})
+
+
+            assert(conservedPcs1.length == conservedPcs2.length)
+            val conservedPcs3 = conservedPcs1
+              .zip(conservedPcs1)
+              .map({case (pcs1, pcs2) => (pcs1 ++ pcs2).distinct})
+
+            val s3 = s1.copy(functionRecorder = functionRecorder3,
+              possibleTriggers = possibleTriggers3,
```
mschwerhoff, line 381: Indentation

mschwerhoff (changes requested):
Looks mostly good, but several comments need to be addressed before the PR can be accepted.

src/main/scala/rules/Consumer.scala
fabianboesiger: `pathConditionAwareMerge` can be used to easily merge the State, it just calls the method `merge` from the State and supplies the corresponding branch conditions. But here, we merge the additional data (I think you called it join data, generically annotated `JD` in the joiner) collected in the branches, for example the partially consumed heap, which is needed again in the continuation.

src/main/scala/rules/Executor.scala
fabianboesiger: That is what I've done originally, but then I discovered a bug where I forgot to supply `joinPoint`, and it was automatically assumed to be `None`. I omitted default initialization with `None` to avoid such bugs in the future, so one has to explicitly use `joinPoint = None`. This makes it more clear in my opinion, but I can change it if you prefer otherwise.

src/main/scala/rules/Consumer.scala
mschwerhoff: Thanks for the technical explanation. What I don't understand, though, is why the additional data is needed here, but not always?

src/main/scala/rules/Executor.scala
mschwerhoff: Nope, all good

src/main/scala/rules/Consumer.scala
fabianboesiger: Maybe I misunderstood your question, but the consumer requires this additional data for the continuation, the producer does not.

The continuation used in the consumer has the following signature:

```
(Q: (State, Heap, Term, Verifier) => VerificationResult)
```

In the producer, the signature of the continuation is:

```
(Q: (State, Verifier) => VerificationResult)
```

The two arguments of type `Heap` and `Term` are the ones that additionally merged in the consumer, in the producer for example we don't need to do this as this additional data simply does not exist in its implementation.

mschwerhoff: Thanks, now it makes sense. I guess I simply haven't taken the time to look at the context.

src/main/scala/rules/Joiner.scala
fabianboesiger: No, `resetState` can be true or false if `moreJoins`is enabled. However, `moreJoins == false ==> resetState == true`.

Intuitively, `resetState` is true for "pure" branches which already are joined without `moreJoins`, and false for impure branches, which are only joined with `moreJoins` enabled.

fabianboesiger:
The benchmarks confirm that this PR does not slow down Silicon compared to the master branch if `moreJoins` is disabled (which it is by default). Two tests however were inconsistent, `allSil/OpenMPSectionReduced-silicon.sil`, `allSil/OpenMPSectionReducedFail1-silicon.sil`, not sure why.

`sbt test` passes all tests.

Issue #579, opened 2021-11-03
Memory leak in Silicon::verify


fpoli:
Calling `Silicon::verify` multiple times (sequentially) on the same verifier instance leaks memory, which is freed only when the verifier is destructed. In other words, there is some state that is not cleaned by Silicon's private `reset` method. In Prusti, this causes our `prusti-server` to go OOM after a while when verifying large programs, because we try to reuse the same verifier instance for multiple verification requests. I'm not sure if the `ViperServer` is affected too.

The memory leak can be debugged in VisualVM by inspecting the diff between heap dumps taken at different times. To given an example, here is one of the paths from the verifier to one of the leaked `viper.silicon.state.terms.Less` instances:
```
this     - value: viper.silicon.state.terms.Less #1
 <- head     - class: scala.collection.immutable.$colon$colon, value: viper.silicon.state.terms.Less #1
  <- equalityDefiningMembers     - class: viper.silicon.state.terms.And, value: scala.collection.immutable.$colon$colon #3843
   <- body     - class: viper.silicon.state.terms.Let, value: viper.silicon.state.terms.And #1
    <- body     - class: viper.silicon.state.terms.Quantification, value: viper.silicon.state.terms.Let #1
     <- [0]     - class: java.lang.Object[], value: viper.silicon.state.terms.Quantification #1
      <- prefix1     - class: scala.collection.immutable.Vector1, value: java.lang.Object[] #6000
       <- postConditionAxioms     - class: viper.silicon.supporters.functions.DefaultFunctionVerificationUnitProvider$functionsSupporter$, value: scala.collection.immutable.Vector1 #16
        <- functionsSupporter$module     - class: viper.silicon.verifier.DefaultMasterVerifier, value: viper.silicon.supporters.functions.DefaultFunctionVerificationUnitProvider$functionsSupporter$ #1
         <- verifier (JNI local)     - class: viper.silicon.Silicon, value: viper.silicon.verifier.DefaultMasterVerifier #1
```

For this particular instance, it seems that `postConditionAxioms` in `functionsSupporter` keeps growing.

fpoli:
Creating a new verifier each time the memory leak *almost* disappears. I can still some non-garbage-collected AST nodes like the following `viper.silver.ast.Seqn`, which seems to be kept alive by the `functionsSupporter` field of the `DefaultMasterVerifier` used in a `VerificationPoolManager`. I suspect there is a (smaller) memory leak there too.

```
[...]

PR #580, opened 2021-11-09
Remove Z3 shutdown hook on stop

fpoli:
This PR should fix the second memory leak identified in #579.


mschwerhoff (approved):
Not sure if the PR is ready to be reviewed, but what's currently here looks good to me

fpoli:
Apparently, `start()` *is* called multiple times without a `stop()` in between.

fpoli:
Unfortunately I cannot easily test this change with the Prusti server to see if the memory leak disappeared; the counterexamples API changed in the meantime. I'll merge this PR anyway as it shouldn't hurt.

PR #583, opened 2021-12-06
Scope for functions, predicates, and methods (of ASTs)

ArquintL:
ViperServer prefers to use the `scope` field of AbstractError to associate errors with particular program members. This PR sets the `scope` field accordingly.


mschwerhoff (approved):
Looks good

PR #588, opened 2022-01-19
Remove quotes around Z3 arguments

fpoli:
The quotation marks that Silicon requires around Z3 arguments are annoying but also unnecessary. Usually, the issue is that scripts don't correctly forward arguments with spaces or quotes, but if so it's the script that should be fixed.

This PR removes the requirement for quotes and fixes the `silicon.sh` script. I also attempted to fix the `silicon.bat` script (`%*` should preserve spaces and quotes and `%CMD%` was probably dropping all quotes), but I don't have a Windows OS to test it.

For reference, I tested the following ways to run Silicon and they all work fine:
* From the sbt shell: `run --z3ConfigArgs "model=true model_validate=true" prog.vpr`
* From bash: `sbt "run --z3ConfigArgs \"model=true model_validate=true\" prog.vpr"`
* From bash: `sbt 'run --z3ConfigArgs "model=true model_validate=true" prog.vpr'`
* From bash: `./silicon.sh --z3ConfigArgs "model=true model_validate=true" prog.vpr`
* From bash: `./silicon.sh --z3ConfigArgs 'model=true model_validate=true' prog.vpr`

This PR is an alternative to https://github.com/viperproject/silicon/pull/568.

This is a (small) breaking change for clients of Silicon that use `--z3Args` or `--z3ConfigArgs`.

### DIFF
--- a/silicon.bat
+++ b/silicon.bat
@@ -26,4 +26,4 @@ cd "%CURR_DIR%"
 
 set JVM_OPTS=-Xss16m -Dlogback.configurationFile="%BASE_DIR%\src\main\resources\logback.xml" -Dfile.encoding=UTF-8
 
-call java "%JVM_OPTS%" -cp "%CP%" viper.silicon.SiliconRunner %*
+call java %JVM_OPTS% -cp "%CP%" viper.silicon.SiliconRunner %*
--- a/silicon.bat
+++ b/silicon.bat
@@ -4,7 +4,7 @@ set CURR_DIR=%cd%
 set BASE_DIR=%~dp0
 
 :: switch to repository root to check for classpath file and possibly call sbt.
-cd %BASE_DIR%
+cd "%BASE_DIR%"
 
 :: Only call sbt if the classpath file is missing.
 if not exist silicon_classpath.txt (
@@ -22,12 +22,8 @@ if not exist silicon_classpath.txt (
 for /f "delims=" %%x in (silicon_classpath.txt) do set CP=%%x
 
 :: switch back to original directory
-cd %CURR_DIR%
+cd "%CURR_DIR%"
 
-set JAVA_EXE=java
-set JVM_OPTS=-Dlogback.configurationFile="%BASE_DIR%\src\main\resources\logback.xml" -Xss16m -Dfile.encoding=UTF-8
-set SILICON_MAIN=viper.silicon.SiliconRunner
-set FWD_ARGS= %*
-set CMD=%JAVA_EXE% %JVM_OPTS% -cp "%CP%" %SILICON_MAIN% %FWD_ARGS%
+set JVM_OPTS=-Xss16m -Dlogback.configurationFile="%BASE_DIR%\src\main\resources\logback.xml" -Dfile.encoding=UTF-8
 
-call %CMD%
+call java "%JVM_OPTS%" -cp "%CP%" viper.silicon.SiliconRunner %*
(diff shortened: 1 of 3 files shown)


fpoli:
By the way, https://www.shellcheck.net/ is great for checking bash scripts.

mschwerhoff:
Thank you! I'll try to find some time soon to test it on Windows, but not before next week.

Did you also test argument forwarding via sbt test, e.g. `testOnly -- -n arithmetic.vpr -Dsilicon:z3ConfigArgs="model=true model_validate=true"`?

fpoli:
> Did you also test argument forwarding via sbt test, e.g. `testOnly -- -n arithmetic.vpr -Dsilicon:z3ConfigArgs="model=true model_validate=true"`?

Is that the exact command? It doesn't seem to work for me, not even on master.

On master, the following works in the sbt shell:
```
sbt:Silicon> testOnly -- -n arithmetic.vpr "-Dsilicon:z3ConfigArgs=\\"model=true model_validate=true\\""
```

With this PR, the previous sbt command incorrectly includes the quote at the beginning of the first key and at the end of the last value :-/ However, the following works fine, but it looks a bit odd:
```
sbt:Silicon> testOnly -- -n arithmetic.vpr "-Dsilicon:z3ConfigArgs=model=true model_validate=true"
```

I'm now thinking that `testOnly` doesn't correctly parse arguments with quotes and spaces. Do you know if it's our code or sbt's code?

mschwerhoff:
> > Did you also test argument forwarding via sbt test, e.g. `testOnly -- -n arithmetic.vpr -Dsilicon:z3ConfigArgs="model=true model_validate=true"`?
> 
> Is that the exact command? It doesn't seem to work for me, not even on master.

I found that snippet in [Silicon's wiki](https://github.com/viperproject/silicon/wiki/Running-Silicon), I'm not surprised that the exact syntax no longer works.

> On master, the following works in the sbt shell:
> 
> ```
> sbt:Silicon> testOnly -- -n arithmetic.vpr "-Dsilicon:z3ConfigArgs=\\"model=true model_validate=true\\""
> ```
> 
> With this PR, the previous sbt command incorrectly includes the quote at the beginning of the first key and at the end of the last value :-/ However, the following works fine (but it looks a bit odd):
> 
> ```
> sbt:Silicon> testOnly -- -n arithmetic.vpr "-Dsilicon:z3ConfigArgs=model=true model_validate=true"
> ```

I can live with that. It's probably a rather esoteric use-case anyway.

> I'm now thinking that `testOnly` doesn't correctly parse arguments with quotes and spaces. Do you know if it's our code or sbt's code?

I guess everybody is involved in this syntax party: `testOnly -- ...` is due to sbt, the `-D...` bit is due to `ScalaTest`, the stuff after `-D` is then handled by us (Silver or Silicon).

fpoli:
The `testOnly -- -n arithmetic.vpr -Dsilicon:z3ConfigArgs="model=true model_validate=true"` that doesn't work in the Sbt shell is a [bug of Sbt](https://github.com/sbt/sbt/issues/6802). It should be fixed in the future Sbt v1.6.3.

fpoli:
Any objection to merging this?

mschwerhoff:
> Any objection to merging this?

Not from my side.

fpoli:
Great; I'll merge. I find really odd that users have to write things like [`--z3Args """""trace=true proof=true"""""`](https://github.com/viperproject/axiom-profiler/issues/36) :)

PR #598, opened 2022-02-09
Check injectivity inhale

tdardinier:
Injectivity is now checked instead of assumed when inhaling quantified permissions.


ArquintL:
Just wanted to do the same fix

PR #599, opened 2022-02-11
Improve position information on reported errors

JonasAlaif:
The idea is to keep around `pvef: ast.Exp => PartialVerificationError` before applying it, so that we can dig down to the actual expression which causes the error. Requires the corresponding PR in silver to be merged.


JonasAlaif:
Actually `ast.Exp => PartialVerificationError` seems to duplicate the functionality of `PartialVerificationError` too much, we'll need to discuss an alternative solution

PR #609, opened 2022-04-22
SMT-LIB 2.6 conformance, support for cvc5, and generalization for alternative SMT-solvers

lfwa:
This pull request contains the following:
- Modified Silicon to only produce SMT-LIB 2.6 conformant encodings (removed Z3-specific statements).
- Support for the [cvc5](https://github.com/cvc5/cvc5) SMT-solver via option `prover`.
- Backend generalized to support alternative SMT-solvers.
- Backwards-compatible config options.

All tests pass with the Z3 solver, but using cvc5 results in ~100 tests failing mostly due to timeouts when using quantifiers.
I will update the PR with benchmarks and a report containing details after 28/04/2022.


README.md
```diff
@@ -134,6 +134,7 @@ method main() {
 * You need recent installations of
   1. the [sbt build tool](https://www.scala-sbt.org/)
   2. the [Z3 SMT solver](https://github.com/Z3Prover/z3/releases)
+  3. the [cvc5 SMT solver](https://github.com/cvc5/cvc5/releases)
```
Aurel300, line 137: Might be good to make it clear that this is optional (or, *one of* the two solvers is required).

src/main/resources/predicate_snap_functions_axioms.smt2
```diff
@@ -32,8 +32,8 @@
 
 (assert (forall ((s $Snap) (pm $PPM)) (!
     ($Perm.isValidVar ($PSF.perm_$PRD$ pm s))
-    :pattern ($PSF.perm_$PRD$ pm s))))
+    :pattern (($PSF.perm_$PRD$ pm s))))) ; ensure :pattern ((f x)) instead of pattern (f x) since pattern must a be a list.
```
Aurel300, line 35: Maybe this comment need not be in all the places this change was done. For future reference it is good to note that this change was made e.g. in the PR description.

src/main/scala/Config.scala
```diff
@@ -185,15 +186,25 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+            + s"not per method or proof obligation (default: 0 i.e. no timeout for all provers except ${Cvc5ProverStdIO.name} with default 180)."),
+    default = None,
     noshort = true
   )
 
+  lazy val timeout: Int = rawTimeout.toOption match {
+    case Some(t) => t
+    // Not set, so compute default value based on prover selected.
+    case _ => prover.toOption match {
+      case Some(Cvc5ProverStdIO.name) => 180
```
Aurel300, line 200: Why? It seems strange to have a different default based on the prover. The description of this in line 191 is also not great.

src/main/scala/Silicon.scala
```diff
@@ -44,10 +44,7 @@ object Silicon {
     s"${BuildInfo.projectVersion}${buildVersion.fold("")(v => s" ($v)")}"
 
   val copyright = "(c) Copyright ETH Zurich 2012 - 2019"
-  val z3ExeEnvironmentVariable = "Z3_EXE"
-  val z3MinVersion = Version("4.5.0")
-  val z3MaxVersion: Option[Version] = None // Some(Version("4.5.0")) /* X.Y.Z if that is the *last supported* version */
-  val dependencies = Seq(SilDefaultDependency("Z3", z3MinVersion.version, "https://github.com/Z3Prover/z3"))
+  val dependencies = Z3ProverStdIO.dependencies ++ Cvc5ProverStdIO.dependencies
```
Aurel300, line 47: What does this do? Similarly to the comment in the README file: is it clear that both of these are not required dependencies?

src/main/scala/decider/Decider.scala
```diff
@@ -87,40 +87,49 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
...
-    def prover: Prover = z3
+    def prover: ProverStdIO = proverStdIO
 
     def pcs: PathConditionStack = pathConditions
 
 //    def setPcs(other: PathConditionStack) = { /* [BRANCH-PARALLELISATION] */
 //      pathConditions = other
 //      pathConditions.assumptions foreach prover.assume
 //    }
+    private def getProverStdIO(prover: String): ProverStdIO = prover match {
```
Aurel300, line 104: Could this be an enum at this point (created when parsing config flags) rather than a `String`?

src/main/scala/decider/SmtlibNameGenerator.scala
```diff
@@ -128,6 +128,65 @@ class SmtlibNameSanitizer extends AbstractNameSanitizer {
     "re.inter", "seq.to.re", "seq.in.re", "re.all", "re.empty",
 
     /* Other reserved words (Z3 specific?) */
-    "min", "List", "const"
+    "min", "List", "const",
+
+    /* cvc5 - Transcendental operators, see https://github.com/cvc5/cvc5/blob/main/src/parser/smt2/smt2.cpp */
+    "exp", "sin", "cos", "tan", "csc", "sec", "cot", "arcsin", "arccos", "arctan", "arccsc", "arcsec", "arccot", "sqrt",
```
Aurel300, line 134: How much of this is CVC5 specific? Are most of these also keywords to Z3? (I see some duplicates.)

src/main/scala/decider/Z3ProverStdIO.scala
```diff
@@ -6,294 +6,43 @@
...
-import viper.silver.reporter.{ConfigurationConfirmation, InternalWarningMessage, Reporter}
+import viper.silver.verifier.{DefaultDependency => SilDefaultDependency}
+import viper.silver.reporter.Reporter
+import viper.silicon.common.config.Version
+
+
+object Z3ProverStdIO {
+  val name = "Z3"
+  val minVersion = Version("4.5.0")
+  val maxVersion = None // Some(Version("4.5.0")) /* X.Y.Z if that is the *last supported* version */
```
Aurel300, line 21: I thought there were actual problems with higher versions @mschwerhoff ?

Aurel300:
The CVC5 support should be tested on the CI also.

src/main/resources/predicate_snap_functions_axioms.smt2
lfwa: Fixed by commit 60de52011848e255b37e077b3a4024f8bfd28a05. The comments are moved into the commit message.

src/main/scala/Config.scala
lfwa: I changed the description in commit 2924e00fec1cc6b264f66840a4ac775ca434e3f3 and added a comment to the file.

The default timeout for cvc5 is set to 180 since it does not support per-checks timeouts set after instantiation of the solver so any checks or saturation checks can cause the verifier to hang indefinitely.
But maybe we just let the default be 0 for all provers and we let cvc5 users deal with this themselves?

I appreciate any input on this :)

src/main/scala/decider/Z3ProverStdIO.scala
lfwa: I cannot offer any input on this since I have not tested it. The `minVersion` and `maxVersion` are just moved from Silicon.scala and the values are kept the same.

src/main/scala/Silicon.scala
lfwa: I moved the dependency definitions from Silicon.scala to the respective ProverStdIO files to avoid cluttering Silicon.scala with further prover-specific information.

I am not entirely sure how to clarify that only the Z3-dependency is needed by default and that the cvc5-dependency is only for when using the cvc5 prover. If you have any ideas, please let me know.

src/main/scala/decider/Decider.scala
lfwa: Letting the factory handle strings instead of anything else was primarily motivated by its simplicity. 

We could create an enum and an extra ValueConverter in Config.scala that converts the strings to an enum but wouldn't we still need a factory that just matches on the enum to create new ProverStdIO objects? To me it seems like this would add another place in the code to update when introducing additional SMT solvers with no added benefits. Admittedly, I am very new to Scala so there are likely advantages that I do not see or a way to easily create a factory with enums.

If you have any suggestions on a clean/nice way to do this, I would be more than happy to introduce the changes if I have time.

src/main/scala/Config.scala
mschwerhoff: I don't have a strong opinion here, but it seems cleaner to not "misuse" the global timeout to compensate for cvc5's lack of per-check timeouts. Please revert to a default of 0 for all provers.

src/main/scala/decider/Z3ProverStdIO.scala
mschwerhoff: @Aurel300  `maxVersion` is currently disabled because no Z3 version really brakes Silicon - whereas Z3 prior to 4.5.0 isn't supported anymore due to configuration option changes. Currently, newer versions make a specify subset of our tests fail (old VerCors examples we suspect of matching loops), but ignoring those, newer versions are supported.

src/main/scala/Silicon.scala
mschwerhoff: If I'm not mistaken, these dependencies are only tracked this way so that Silver can report them, not sure if that feature has ever been used, though. I also don't think that the mechanism can differentiate between mandatory and optional dependencies, or dependencies that depend on command-line options.

@lfwa No need for further changes.

src/main/scala/decider/Decider.scala
mschwerhoff: Enums (or sealed traits) have the advantage over strings that the compiler can perform exhaustiveness checks for the former. It is therefore good practice to turn strings into enums as early as possible (i.e. in the command-line parser). Another disadvantage of strings is that you always need to account for upper-/lowercase spelling.

Aurel300:
@lfwa The CI file is [here](https://github.com/viperproject/silicon/blob/master/.github/workflows/ci.yml). Z3 is installed as part of the base "Gobra" container, I think. I don't think this includes CVC5, so it would have to be installed manually. Then, the whole [test suite](https://github.com/viperproject/silicon/blob/master/.github/workflows/ci.yml#L63) should be run with the `--prover cvc5` algorithm passed (somehow, not sure what the syntax is for `sbt test`).

src/main/scala/Config.scala
lfwa: Fixed/reverted back to default of 0 for all provers in 37e3adb6f12fbd620710edc401ecd78ef87b5bb7, 6ab985857e5f8dafa1e3645d6b110d8075e7bb00, and 4215b09691a7a66163388c4e562b5b00f9941c7e.

src/main/scala/decider/Decider.scala
lfwa: Great, thanks for the explanation.
I will introduce enums instead of the string factory if I have time. :)

PR #620, opened 2022-06-03
Don't count expected errors towards `numberOfErrorsToReport` limit

JonasAlaif:
Linked to https://github.com/viperproject/silver/pull/583


mschwerhoff (approved):
I like concise PRs :-)

JonasAlaif:
@mschwerhoff It seems that [`errorsReportedSoFar`](https://github.com/viperproject/silicon/blob/refute/src/main/scala/verifier/Verifier.scala#L43) doesn't get reset between test runs, and so causes the refutation tests to fail. We should reset this number to 0 between tests; ideally somewhere [here](https://github.com/viperproject/silicon/blob/refute/src/test/scala/SiliconTests.scala#L34). Any idea how I can get access to the `errorsReportedSoFar` val from there?

mschwerhoff:
Not your fault, of course, but `Verifier` is not a `StatefulComponent` and should not have state that needs to be reset. `BaseVerifier` would probably be a better fit. Resetting should then happen in `def reset()`.

PR #622, opened 2022-06-08
Improve SymbExLogger

niomaster:
A first attempt at some changes to symbexlogger I'd like to see:
- `SymbLog` doesn't write records into memory directly, but rather does so via a `SymbLogListener`, so VerCors can listen to the events directly instead of polling the log. Additionaly I don't think we need the history of all records.
- I made `SymbLog` thread safe up to parallelization of entities, as far as I understand that's the current state of silicon (?). If this can be left in I'd be inclined to turn on logging by default if the performance penalty is not too great.

Of course happy to address any comments :)


src/main/scala/decider/ProverStdIO.scala
```diff
@@ -347,6 +348,45 @@ abstract class ProverStdIO(uniqueId: String,
     toMap(stats)
   }
 
+  def statistics(): Map[String, String] = {
+    lastStatistics = queryStatistics()
+    lastStatistics
+  }
+
+  /**
```
mschwerhoff, line 356: Could this delta computation be moved out of ProverStdIO? The statistics come from the prover, but their postprocessing is not tightly coupled to proving assertions.

mschwerhoff:
Regarding thread safety: Generally a good idea, but it seems a bit dangerous to make one logging/reporting component thread-safe, while many others still are not. I don't have a strong opinion here, though, and since you already did the work ... :-)

mschwerhoff:
Is this work in progress or should it be reviewed? In the former case, please convert to a draft (I've actually never used that GitHub feature, but since it exists ...).

niomaster:
This is ready for review! Sorry, did not know if I should re-ping you.

mschwerhoff:
Re-pinging is always a good idea :-) I'll do a review after my holiday, i.e. mid/end August.

mschwerhoff:
@niomaster Please remove the merge conflicts first.

Also: why is a ``validateOpt(...)` remove from file `Config.scala`? Is that change from you?

niomaster:
I resolved the merge conflicts, and the `validateOpt` bit I removed checks that ideModeAdvanced ==> numberOfParallelVerifiers = 1, which is no longer necessary, since SymbExLogger is thread safe.

mschwerhoff (approved):
Looks good to me.

@ArquintL If you see a reason not to merge, please let me know. Otherwise, I'll merge at the end of the week.

src/main/scala/logger/LogConfig.scala
```diff
@@ -7,10 +7,23 @@
...
 case class LogConfig(isBlackList: Boolean,
                      includeStore: Boolean, includeHeap: Boolean, includeOldHeap: Boolean, includePcs: Boolean,
-                     recordConfigs: List[RecordConfig])
+                     recordConfigs: List[RecordConfig]) {
+  def getRecordConfig(d: DataRecord): Option[RecordConfig] = {
+    for (rc <- recordConfigs) {
+      if (rc.kind.equals(d.toTypeString)) {
+        rc.value match {
+          case Some(value) => if (value.equals(d.toSimpleString)) return Some(rc)
+          case _ => return Some(rc)
```
ArquintL, line 20: This case looks a bit odd. Should this be `None`?

ArquintL (approved):
I only have a minor comment but otherwise looks good to me too

niomaster:
Just in case: I'm working on making the logger externally parallelizable (as is silicon now) as promised in <https://github.com/viperproject/silicon/issues/660#issuecomment-1316983192>, so please don't merge this right now.

niomaster:
Quick question about this bit of code: <https://github.com/niomaster/silicon/blob/aa574464347a72eafdac6315e3ffc11e1bba7d2f/src/main/scala/Silicon.scala#L195-L232>

In the time out code the log is printed after all, but is it at that point guaranteed that the verification run is terminated? If not, I think it's not thread-safe to render the log while it's still potentially being altered (even with one silicon instance and `--numberOfParallelVerifiers 1`). If I'm reading correctly even `shutdown` + `shutdownNow` do not guarantee that actively running tasks have terminated, so that's also not a suitable point to print the log? If you don't want to bother with the java interrupt logic business I can probably also just make it so the log starts ignoring further records, that's an option. Thanks!

ArquintL:
> Quick question about this bit of code: https://github.com/niomaster/silicon/blob/aa574464347a72eafdac6315e3ffc11e1bba7d2f/src/main/scala/Silicon.scala#L195-L232
> 
> In the time out code the log is printed after all, but is it at that point guaranteed that the verification run is terminated? If not, I think it's not thread-safe to render the log while it's still potentially being altered (even with one silicon instance and `--numberOfParallelVerifiers 1`). If I'm reading correctly even `shutdown` + `shutdownNow` do not guarantee that actively running tasks have terminated, so that's also not a suitable point to print the log? If you don't want to bother with the java interrupt logic business I can probably also just make it so the log starts ignoring further records, that's an option. Thanks!

The original idea was to get data even in the case that verification fails (exceptionally). Your suggestion of providing a log in this case but ignoring all further entries seems very reasonable to me as this should provide all information one needs to (hopefully) figure out the cause of the failure

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -192,36 +185,42 @@ import scala.util.{Failure, Success, Try}
...
 
-  def getRecordConfig(d: DataRecord): Option[RecordConfig] = {
-    for (rc <- logConfig.recordConfigs) {
-      if (rc.kind.equals(d.toTypeString)) {
-        rc.value match {
-          case Some(value) => if (value.equals(d.toSimpleString)) return Some(rc)
-          case _ => return Some(rc)
-        }
-      }
+  private def parseLogConfig(c: Config): LogConfig = {
```
mschwerhoff, line 196: Please emit/log warnings if a config parsing step fails

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -192,36 +185,42 @@ import scala.util.{Failure, Success, Try}
...
+  protected var members: immutable.Map[ast.Member, Log] = Map.empty
+  protected var closed: Boolean = false
+  def isClosed: Boolean = closed
+  protected val uidCounter: AtomicInteger = new AtomicInteger(0)
+
+  protected def newEntityLogger(member: ast.Member, pcs: PathConditionStack): Log
+
+  def freshUid(): Int = uidCounter.getAndIncrement()
+
+  def whenEnabled(f: => Unit): Unit = f
```
mschwerhoff, line 221: Would anything change if we made the closure explicit: `f: () => Unit` and `... = f()`? I find the current Scala syntax very unintuitive, in particular for people that don't know Scala very well.

Just `f()` would already be more explicit.

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -232,256 +231,177 @@ object SymbExLogger {
...
-    */
-  def currentLog(): SymbLog = {
-    if (enabled)
-      memberList.last
-    else NoopSymbLog
+      log.openMemberScope()
+      log
+    } catch {
+      case anything: Throwable =>
+        println(anything)
```
mschwerhoff, line 252: Silicon uses a proper logging framework. Please use that instead.

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -232,256 +231,177 @@ object SymbExLogger {
...
+      closed = true
+    }
 
-  private def appendLog(r: SymbolicRecord, ignoreBranchingStack: Boolean = false): Unit = {
-    if (isClosed) {
-      logger warn "ignoring record insertion to an already closed SymbLog instance"
-      return
+  def whenOpen(f: => Unit): Unit =
+    synchronized {
+      if(!isClosed) f else ()
```
mschwerhoff, line 320: `f()`?

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -545,22 +448,143 @@ class SymbLog(v: ast.Member, s: State, pcs: PathConditionStack) {
...
+    }
+    whenEnabled { lastStatistics = lastStatistics ++ currentStatistics }
+    currentStatistics ++ deltaStatistics
+  }
+
+  private def getDelta(pair: (String, String)): (String, Option[String]) = {
+    val curValInt = pair._2.toIntOption
+    val prevValInt = lastStatistics.get(pair._1) match {
+      case Some(value) => value.toIntOption
+      case _ => Some(0) // value not found
```
mschwerhoff, line 471: Magic constants are dangerous. Can't `None` be returned instead?

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -545,22 +448,143 @@ class SymbLog(v: ast.Member, s: State, pcs: PathConditionStack) {
...
+    }
+    val curValDouble = pair._2.toDoubleOption
+    val prevValDouble = lastStatistics.get(pair._1) match {
+      case Some(value) => value.toDoubleOption
+      case _ => Some(0.0) // value not found
```
mschwerhoff, line 476: See preceding comment

src/main/scala/logger/SymbExLogger.scala
```diff
@@ -192,36 +185,42 @@ import scala.util.{Failure, Success, Try}
...
-
-  var enabled = false
-  var logConfig: LogConfig = LogConfig.default()
-
-  def freshUid(): Int = {
-    val uid = uidCounter
-    uidCounter = uidCounter + 1
-    uid
+case object SymbExLogger {
+  def ofConfig(c: Config): SymbExLogger[_ <: MemberSymbExLogger] = {
```
mschwerhoff, line 189: It took me a while to realise that `c` is Silicon's config, and `config` further down (and in other files, e.g. `DataRecord.scala` ) is the symbex-logger config. Since `config` everywhere else in Silicon denotes Silicon's configuration object, it would be best to not overload the name. If it's not too much work, please rename the symbex-logger config object to, e.g. `symbExConfig`  or `loggerConfig`.

src/main/scala/logger/records/data/FunctionRecord.scala
```diff
@@ -12,9 +12,8 @@ import viper.silicon.state.State
 import viper.silicon.state.terms.Term
 import viper.silver.ast
 
-class FunctionRecord(v: ast.Function, s: State, p: PathConditionStack) extends MemberRecord {
+class FunctionRecord(v: ast.Function, p: PathConditionStack) extends MemberRecord {
```
mschwerhoff, line 15: Why is the `State` (here and further down) no longer relevant?

src/main/scala/rules/Brancher.scala
```diff
@@ -100,8 +100,8 @@ object brancher extends BranchingRules {
         }
 
         (v0: Verifier) => {
-          SymbExLogger.currentLog().switchToNextBranch(uidBranchPoint)
-          SymbExLogger.currentLog().markReachable(uidBranchPoint)
+          v.symbExLog.switchToNextBranch(uidBranchPoint)
```
mschwerhoff, line 103: `v0` instead of `v`?

src/main/scala/rules/Consumer.scala
```diff
@@ -158,10 +158,10 @@ object consumer extends ConsumptionRules {
       val h0 = s0.h /* h0 is h, but potentially consolidated */
       val s1 = s0.copy(h = s.h) /* s1 is s, but the retrying flag might be set */
 
-      val sepIdentifier = SymbExLogger.currentLog().openScope(new ConsumeRecord(a, s1, v.decider.pcs))
+      val sepIdentifier = v.symbExLog.openScope(new ConsumeRecord(a, s1, v.decider.pcs))
 
       consumeTlc(s1, h0, a, pve, v1)((s2, h2, snap2, v2) => {
-        SymbExLogger.currentLog().closeScope(sepIdentifier)
+        v.symbExLog.closeScope(sepIdentifier)
```
mschwerhoff, line 164: `v2`?

src/main/scala/rules/Consumer.scala
```diff
@@ -186,31 +186,31 @@ object consumer extends ConsumptionRules {
...
       case imp @ ast.Implies(e0, a0) if !a.isPure =>
         val impliesRecord = new ImpliesRecord(imp, s, v.decider.pcs, "consume")
-        val uidImplies = SymbExLogger.currentLog().openScope(impliesRecord)
+        val uidImplies = v.symbExLog.openScope(impliesRecord)
 
         evaluator.eval(s, e0, pve, v)((s1, t0, v1) =>
           branch(s1, t0, Some(e0), v1)(
             (s2, v2) => consumeR(s2, h, a0, pve, v2)((s3, h1, t1, v3) => {
-              SymbExLogger.currentLog().closeScope(uidImplies)
+              v.symbExLog.closeScope(uidImplies)
```
mschwerhoff, line 194: `v3`?

Potentially also applies to additional changes further down.

src/main/scala/verifier/WorkerVerifier.scala
```diff
@@ -7,6 +7,7 @@
 package viper.silicon.verifier
 
 import viper.silicon.supporters._
+import viper.silver.ast.Member
```
mschwerhoff, line 10: To avoid confusing Silver AST nodes with Silicon terms nodes, always use AST members as `ast.Foo` in contexts where AST nodes and terms might both be used.

mschwerhoff:
@niomaster FYI: We just started preparing the Januar release of Viper, and would be happy to include your changes.

src/main/scala/logger/SymbExLogger.scala
niomaster: Yes, it would change at the usage site, e.g.:

```
def defn(f: => Unit): Unit = f

def use(): Unit = {
  defn {
    println()
  }
}
```

becomes e.g.:

```
def defn(f: () => Unit): Unit = f()

def use(): Unit = {
  defn { () =>
    println()
  }
}
```

I agree that it's a fairly arcane scala feature. I sometimes use it when the usage of it is intuitive (like I think it is here), at the cost of a weird definition. Happy to make it more explicit, though.

src/main/scala/logger/SymbExLogger.scala
niomaster: Oh this is just a debugging remnant, I don't think there is a concrete reason to catch or log exceptions here.

src/main/scala/logger/records/data/FunctionRecord.scala
niomaster: I have no opinion on whether it's useful, but currently it was set to `null` in all paths.

src/main/scala/rules/Brancher.scala
niomaster: My current understanding is that the verifier is carried through continuations only to enable branch parallelization, is that the case? If `v0 != v`, then the `symbExLog` of `v0` will be an incorrect `MemberSymbExLogger`, since the current member is accounted within one verifier. If there are cases when `v0 != v` other than when branch parallelization is enabled I guess we need some way to deal with that.

src/main/scala/Config.scala
```diff
@@ -754,19 +754,15 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
-  validateOpt(ideModeAdvanced, numberOfParallelVerifiers) {
-    case (Some(false), _) =>
-      Right(())
-    case (Some(true), Some(n)) =>
-      if (n == 1)
-        Right(())
-      else
-        Left(  s"Option ${ideModeAdvanced.name} requires setting "
-             + s"${numberOfParallelVerifiers.name} to 1")
+  validateOpt(ideModeAdvanced, parallelizeBranches) {
```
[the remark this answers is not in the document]
niomaster: How about just `enableSymbExLog`? Also I think the actual option is defined in the silver repo, so do I file an additional PR there? Do feel free to just rename it yourself and be done with it :)

src/main/scala/logger/SymbExLogger.scala
mschwerhoff: Thanks, I retract my request. The syntax overhead at call-site is not worth the additional explicitness at definition-site.

src/main/scala/rules/Brancher.scala
mschwerhoff: Parallelisation is currently indeed the only motivation for threading the verifier through. But others are generally possible, e.g. instead of using state flags that control certain operations (such as extended-exhale), one could temporarily swap out algorithms, and these could be attached to the verifier.

But that's all hypothetical. It would be good, however, to document why `v` is used here, and what the expected relation between `v` and `v0` is.

src/main/scala/Config.scala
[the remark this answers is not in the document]
mschwerhoff: Good point about the Silver repo. No worries, then leave as-is, and we'll rename it ... eventually :-)

src/main/scala/rules/Brancher.scala
niomaster: Taking that into account and after thinking about it a bit more I think `v0` is less wrong. Using `v` gives the impression that we will always use the original verifier for symbExLogging, but in fact we only delay the problem by ~1 log record if e.g. the verifier is (sequentially) swapped out. Consistently using the current verifier at least gives options, where you can exclude symbExLogging with the option (as with branch parallelization) or manually account the log in subverifiers.

src/main/scala/logger/SymbExLogger.scala
niomaster: Yes, the `-delta` key will be missing from the first statistics, but that makes sense.

niomaster:
I've addressed all the comments!

mschwerhoff (approved):
Thank you!

marcoeilers:
No objections, looks good to me.

mschwerhoff:
@niomaster Thanks a lot for your contribution to Viper, Pieter!

Issue #630, opened 2022-07-07
Silicon verifies example that should fail.


zgrannan:
The following program verifies in Silicon, but it shouldn't. In Carbon, verification fails as expected.

```
field discriminant: Int

field field1: Ref

predicate pred(self: Ref) {
  acc(self.discriminant, write) &&
  acc(self.field1, write) &&
  (self.discriminant == 4 ?
    predFive(self.field1) :
    (false ? tt(self.field1) : true))
}

predicate predFive(self: Ref) {
  true && true
}

predicate tt(self: Ref) {
  true
}

method m_test1() returns ()
{
  var _1: Ref
  inhale acc(pred(_1), write)
  inhale (unfolding acc(pred(_1), write) in _1.discriminant == 4 &&
    (unfolding predFive(_1.field1) in true))
  assert false
}
```

Interestingly, replacing `true && true` with `true` in `predFive` causes the verifier to (correctly) report an error.

Also, replacing `false ? tt(self.field1) : true` with `true` causes the verifier to (correctly) report an error.

This is a minimal case from a more complex example (in the real example `predFive` refers to `self`)

PR #632, opened 2022-07-12
Fix encoding of path conditions as implications

zgrannan:
Currently when path conditions are encoded as implications, the previous path conditions do not appear LHS of the implication. For example, consider a path`P0` with assertion `A`  followed by `P1` with assertion `B`. The currently generated assertions are `P0 => A` and `P1 => B`. However, if the assertion `B` was generated due to being in the `P0` branch, then this is unsound (consider, for example, if `P1` is `True`; then `P1 => B` is just `B`).

This PR changes the behaviour to include the previous path conditions on the LHS; such that the assertion for P1 would be `(P0 && P1) => B` instead of `P1 => B`.

Fixes https://github.com/viperproject/silicon/issues/630.

### DIFF
--- a/src/main/scala/decider/PathConditions.scala
+++ b/src/main/scala/decider/PathConditions.scala
@@ -147,8 +147,10 @@ private trait LayeredPathConditionStackLike {
     for (layer <- layers.reverseIterator) {
       unconditionalTerms ++= layer.globalAssumptions
 
-      layer.branchCondition foreach { condition =>
-         implicationLHS = And(implicationLHS, condition)
+      layer.branchCondition match {
+        case Some(condition) =>
+          implicationLHS = And(implicationLHS, condition)
+        case None =>
       }
 
       conditionalTerms :+=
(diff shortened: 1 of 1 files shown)


src/main/scala/decider/PathConditions.scala
```diff
@@ -142,12 +142,17 @@ private trait LayeredPathConditionStackLike {
...
     var unconditionalTerms = Vector.empty[Term]
     var conditionalTerms = Vector.empty[Term]
+    var implicationLHS: Term = True()
 
-    for (layer <- layers) {
+    for (layer <- layers.reverseIterator) {
       unconditionalTerms ++= layer.globalAssumptions
 
+      layer.branchCondition foreach { condition =>
+         implicationLHS = And(implicationLHS, condition)
```
mschwerhoff, line 151: `And` is n-ary - can't that be exploited to avoid the loop over layer.branchCondition?

zgrannan: @mschwerhoff `layer.branchCondition` is an `Option[Term]` so that inner block will execute at most once

zgrannan: Yeah that's probably more clear, I've changed it to use a match instead.

Aurel300:
Looks good. @JonasAlaif is this the way to update the `silver` submodule? Should there be a separate PR in that repo?

JonasAlaif:
@Aurel300/@zgrannan you want to merge this first, but you'll also need to create a PR in silver and merge the changes there too

PR #519, opened 2020-11-06
Use Z3's API

fpoli:
Do not merge. This PR is a proof of concept for using Z3's API from Silicon. On a few Viper examples that I tried it seems to work fine, with up to a 2x slowdown (e.g. 58s -> 2m). This is expected as Silicon is still pretty-printing all the commands as before, so there is room for optimizations.
* The `Z3_EXE` variable is no longer used.
* The code has not been optimized, and is e.g. quadratic in the number of sort and function definitions.
* Various hacks are used to still support clients who want to directly emit SMT-LIB2 commands (e.g. in a case we pretty print a term, match it with a regex, parse it with a SMT-LIB2 parser, use it to generate another SMT-LIB2 command, and finally pass the command to Z3, which will use its own parser).
* The PR overwrites the existing `Z3ProverStdIO` just to have a clear diff of the modifications.

What this PR doesn't do:
* [ ] Move the modifications to a new `Z3ProverLib` class, adding a command line option to choose between it and the usual `Z3ProverStdIO`.
* [ ] Rewrite clients of `Z3ProverStdIO` to rely less on `emit(String)` and more on a new `emitTerm(Term)` method.
* [ ] Implement a conversion from `Term` to Z3's AST without going through the SMT-LIB2 format. This will allow to remove hacks and conversions of `Term` types to `String`.
* [ ] Compile Z3 enabling the Java API. This means that the Z3 artifacts produced by Jenkins should include e.g. the `libz3.so`, `libz3java.so` and `z3.jar` files.

PR #633, opened 2022-07-26
Interaction with Z3 through Z3's Java API instead of StdIO

marcoeilers:
Requires libz3java to be on the LD_LIBRARY_PATH (or the Windows equivalent) or the java.library.path. The included Java files are from Z3 version 4.8.6, so the library should have the same version, but apparently 4.8.7 also works. Newer versions do not.

Some interesting points:
 - Silicon emits lots of function declarations and axioms for various stuff as SMT2LIB strin gbefore verifying stuff. We parse this, but to be able to do this, we need to parse this stuff all at once. So we collect all of this stuff and then parse it all at once (once the preamble is over).
 - We pass in settings directly via the API and just ignore settings passed as set-option strings; as a side note, half of the options we use no longer exist in the current Z3 version, and passing non-existing options via API results in an exception.
 - We cache function declarations and sort translations.
 - Counterexample (model) extraction via API is implemented, too.
 - The entire test suite should work, except that there is an example that contains invalid triggers; while this results in a warning when using Z3 via StdIO, it results in an exception when used via API, so that example won't verify.
 - To handle built-in SMTLIB functions not directly available in Viper (e.g., bitvectors and floats), we use a bit of a hack, since we cannot use them via API from only a string (they have their own API functions). So we create mini-SMTLIB queries that use those functions, let Z3 parse them, and take what we need from the generated ASTs.

Probably shouldn't be merged before the release.
I measured a 10-40% speedup for examples from the test suite and much higher speedups for some very long running examples.


Aurel300:
(Note on macOS the equivalent is `DYLD_LIBRARY_PATH`.)

src/main/scala/decider/TermToZ3APIConverter.scala
```diff
@@ -0,0 +1,525 @@
...
+      case NoPerm() => ctx.mkReal(0)
+      case FractionPermLiteral(r) => ctx.mkDiv(convertToReal(IntLiteral(r.numerator)), convertToReal(IntLiteral(r.denominator)))
+      case FractionPerm(n, d) => ctx.mkDiv(convertToReal(n), convertToReal(d))
+      case PermLess(t0, t1) => ctx.mkLt(convertTerm(t0).asInstanceOf[ArithExpr], convertTerm(t1).asInstanceOf[ArithExpr])
+      case PermAtMost(t0, t1) => ctx.mkLe(convertTerm(t0).asInstanceOf[ArithExpr], convertTerm(t1).asInstanceOf[ArithExpr])
+      case PermPlus(t0, t1) => ctx.mkAdd(convertTerm(t0).asInstanceOf[ArithExpr], convertTerm(t1).asInstanceOf[ArithExpr])
+      case PermMinus(t0, t1) => ctx.mkSub(convertTerm(t0).asInstanceOf[ArithExpr], convertTerm(t1).asInstanceOf[ArithExpr])
+      case PermTimes(t0, t1) => ctx.mkMul(convertTerm(t0).asInstanceOf[ArithExpr], convertTerm(t1).asInstanceOf[ArithExpr])
+      case IntPermTimes(t0, t1) => ctx.mkMul(convertTerm(t0).asInstanceOf[ArithExpr], convertTerm(t1).asInstanceOf[ArithExpr])
+      case PermIntDiv(t0, t1) => ctx.mkDiv(convertToReal(t0), convertToReal(t1))
```
Aurel300, line 319: Is there no int division in Z3?

src/main/scala/decider/TermToZ3APIConverter.scala
```diff
@@ -0,0 +1,525 @@
...
+    funcDeclCache.update((name, argSorts, resSort), res)
+    res
+  }
+
+
+  @inline
+  protected def createSMTApp(functionName: String, args: Seq[Term], outSort: Sort): Z3Expr = {
+    // workaround: since we cannot create a function application with just the name, we let Z3 parse
+    // a string that uses the function, take the AST, and get the func decl from there, so that we can
+    // programmatically create a func app.
```
Aurel300, line 462: Is this because we don't know the types for the arguments? Are we passing them to Z3 so it can figure out their sorts?

src/main/scala/decider/TermToZ3APIConverter.scala
```diff
@@ -0,0 +1,525 @@
...
+    val decls = args.zipWithIndex.map{case (a, i) => s"(declare-const workaround${i} ${smtlibConverter.convert(a.sort)})"}.mkString(" ")
+    val funcAppString = s"(${functionName} ${(0 until args.length).map(i => "workaround" + i).mkString(" ")})"
+    val assertion = decls + s" (assert (= ${funcAppString} ${funcAppString}))"
+    val workaround = ctx.parseSMTLIB2String(assertion, null, null, null, null)
+    val app = workaround(0).getArgs()(0)
+    val decl = app.getFuncDecl
+    val actualArgs =  if (decl.getArity > args.length){
+      // the function name we got wasn't just a function name but also contained a first argument.
+      // this happens with float operations where functionName contains a rounding mode.
```
Aurel300, line 471: Can this happen for other reasons (e.g. accidentally calling a function with the wrong number of arguments)? Could you whitelist only float-related operations in this branch?

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,439 @@
...
+  val name = "Z3-API"
+  val minVersion = Version("4.8.6.0")
+  val maxVersion = Some(Version("4.8.6.0")) /* X.Y.Z if that is the *last supported* version */
+  val exeEnvironmentalVariable = "Z3_EXE"
+  val dependencies = Seq(SilDefaultDependency("Z3", minVersion.version, "https://github.com/Z3Prover/z3"))
+  val staticPreamble = "/z3config.smt2"
+  val startUpArgs = Seq("-smt2", "-in")
+  val randomizeSeedsOptions = Seq("random_seed")
+  val randomizeSeedsSetting = "RANDOMIZE_SEEDS"
+  val initialOptions = Map("auto_config" -> "false", "type_check" -> "true")
```
Aurel300, line 40: Would be good to comment what the various flags do for future reference? Do we know, or are they copied over from an older codebase?

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,439 @@
...
+      emittedSortSymbols.clear()
+    }
+  }
+
+  def push(n: Int = 1): Unit = {
+    endPreamblePhase()
+    pushPopScopeDepth += n
+    if (n != 1)
+      sys.error("Not supported")
+    prover.push()
```
Aurel300, line 160: I assume `n > 1` is simply not used, although I guess we could have just had a loop here. Still, could the `n` arguments simply be removed everywhere?

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,439 @@
...
+    // we ignore this, don't know any better solution atm.
+    // our settings are defined in this class (see above).
+    // except we check if someone gives us our custom randomization string, in which
+    // case we'll directly set the options to randomize.
+    val optionSetString = s"(set-option :${Z3ProverAPI.randomizeSeedsSetting}"
+    if (contents.exists(s => s.startsWith(optionSetString))) {
+      val params = ctx.mkParams()
+      Z3ProverAPI.randomizeSeedsOptions.foreach(o => params.add(o, Random.nextInt(10000)))
+      prover.setParameters(params)
+    }
```
Aurel300, line 193: Can we emit warnings for ignored settings?

src/main/scala/decider/TermToZ3APIConverter.scala
marcoeilers: No, the problem is that there is nothing like ctx.getInterpretedFuncDecl(name) in the API that I could give a string like "fp.add" to and get the declaration for floating point addition. Instead, for any built-in interpreted function, there is a separate creator method in the API (e.g., ctx.mkFloatAdd(arg1, arg2)). So I'd have to have a big match-case here that matches all names of all built-in functions and then calls the respectice API method. Which I didn't want to do now because there are a lot of them, and the way we currently use SMTFuncApps, sometimes, the strings aren't actually *just* a function name but a function name and some number of default argument (e.g., "fp.add RNE", which contains the function name and a first argument that determines the rounding mode).

src/main/scala/decider/TermToZ3APIConverter.scala
marcoeilers: There is, line 296. These are just the perm arithmetic operations.

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,439 @@
...
+    val statistics = prover.getStatistics
+    val result = mutable.HashMap[String, String]()
+    for (e <- statistics.getEntries()) {
+      result.update(e.Key, e.getValueString)
+    }
+    ListMap.from(result)
+  }
+
+  def comment(str: String): Unit = {
+    if (!preamblePhaseOver && str == "End preamble")
```
Aurel300, line 316: Is this `str == ...` check ever necessary to detect the end of preamble?

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,439 @@
...
+    val fun = Fun(id, argSorts, resultSort)
+    val decl = FunctionDecl(fun)
+
+    termConverter.convert(decl)
+
+    fun
+  }
+
+  def declare(decl: Decl): Unit = {
+    // just creating the term is enough.
```
Aurel300, line 331: What is this comment referring to? The declaration is still added to `emitted*`, no?

src/main/scala/decider/TermToZ3APIConverter.scala
marcoeilers: I guess it could, yes. 

I think if we want to avoid this special case (which really shouldn't be here), we should just forbid including arguments in the function name in general, then we wouldn't have to handle this case here. It's just that that was unproblematic when we were dealing only with strings, so I did it when making the factory for floating point operations. If we rewrite that code,  I think the only consequence is that users of floating point functions would always have to pass the rounding mode explicitly.

src/main/scala/decider/Z3ProverAPI.scala
marcoeilers: They are copied from the Z3 options in z3config.smt2, and I removed the options that no longer work (since they result in ugly error messages). I'll add a comment.

src/main/scala/decider/Z3ProverAPI.scala
marcoeilers: I haven't seen it used anywhere, but there might be a special case that does (Malte would know I guess). I'll just add a loop though, the not supported case is a leftover from when I first tried this out and just wanted to get it running, and I just forgot to fix it.

src/main/scala/decider/Z3ProverAPI.scala
marcoeilers: Z3 will output an error and potentially an exception when you give it settings that it doesn't recognize via API.

src/main/scala/decider/Z3ProverAPI.scala
marcoeilers: When calling ctx.mkFuncDecl, the declaration is added to Z3's current state, without us having to explicitly call another function (e.g., ctx.declareFunc(ctx.mkFuncDecl(...)). That's what I meant there, will clarify.

src/main/scala/decider/Z3ProverAPI.scala
marcoeilers: Apparently not, I'll remove it.

fpoli:
I don't remember if there is anything useful, but you can also check https://github.com/viperproject/silicon/pull/519.

Aurel300:
One particular difference I noticed is using a "managed" library for the Z3 APIs in `build.sbt`, along with some updates to the `silicon.sh` script to accommodate for this. @marcoeilers is this the same library you are using here? Having a managed dependency is probably better than shipping a JAR like this.

marcoeilers:
Oh, I only saw your comments just now. Yes, I agree a proper library dependency is much nicer.

marcoeilers:
As far as I'm concerned, since the release is done now, we could merge this. @mschwerhoff what do you think?

marcoeilers:
Merged as part of the parallel branch verification PR.

PR #634, opened 2022-08-11
Parallel branch verification

marcoeilers:
(Depends on the Z3 API PR (not because there are real functional dependencies but simply because I merged the two at some point to be able to test both together))

Building on Malte's earlier attempt to get parallel branch verification, this PR adds a command line parameter --parallelizeBranches that directs Silicon to verify different branches in parallel.
We use a ForkJoinPool for this which uses a work stealing algorithm, so every worker has its own Z3 instance and its own work queue. When it reaches a branch, it puts the else-branch into its own queue s.t. other idle workers can take it; if noone does, it will take the else-branch-task from the beginning of its own queue and verify it itself.
Workers that become idle will steal from the *end* of the queue of another worker and therefore get the oldest task that worker still has left over, which tends to be the biggest task in the queue, which should avoid switching back and forth between workers for tiny tasks.
Every task switch requires updating the worker's solver to the branch's current state, i.e., popping all its current path conditions and pushing all path conditions from the task it's taking over. 

Speedups obviously depend on the program and the machine, but if you have a bunch of cores lying around and your program has lots of branching, things will become MUCH faster (e.g. 6x for Nagini VerifyThis examples on my desktop at home).


src/main/scala/decider/Decider.scala
```diff
@@ -87,23 +89,33 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
   def identifierFactory: IdentifierFactory
 
   object decider extends Decider with StatefulComponent {
-    private var proverStdIO: ProverStdIO = _
+    private var proverStdIO: Prover = _
```
mschwerhoff, line 92: Is `proverStdIO` still appropriate? Rename to just `prover`?

src/main/scala/decider/Decider.scala
```diff
@@ -87,23 +89,33 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
...
-//      pathConditions.assumptions foreach prover.assume
-//    }
-    private def getProverStdIO(prover: String): ProverStdIO = prover match {
+    def setPcs(other: PathConditionStack) = {
+      /* [BRANCH-PARALLELISATION] */
+      pathConditions = other
+      while (prover.pushPopScopeDepth > 1){
+        prover.pop()
+      }
+      val layeredStack = other.asInstanceOf[LayeredPathConditionStack]
```
mschwerhoff, line 108: The downcast makes me think that the interface should be changed. Maybe add as TODO.

src/main/scala/decider/PathConditions.scala
```diff
@@ -221,7 +221,7 @@ private[decider] class LayeredPathConditionStack
        with PathConditionStack
        with Cloneable {
 
-  private var layers: Stack[PathConditionStackLayer] = Stack.empty
+  var layers: Stack[PathConditionStackLayer] = Stack.empty
```
mschwerhoff, line 224: Please keep `/* private */` to indicate that the field should not be accessed, or at least not modified, from the outside.

src/main/scala/decider/ProverStdIO.scala
```diff
@@ -30,7 +31,7 @@ abstract class ProverStdIO(uniqueId: String,
     extends Prover
        with LazyLogging {
 
-  protected var pushPopScopeDepth = 0
+  var pushPopScopeDepth = 0
```
mschwerhoff, line 34: Please keep `/* protected */`

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,457 @@
...
+
+class Z3ProverAPI(uniqueId: String,
+                  termConverter: TermToZ3APIConverter,
+                  identifierFactory: IdentifierFactory,
+                  reporter: Reporter)
+    extends Prover
+      with LazyLogging
+{
+
+  var pushPopScopeDepth = 0
```
mschwerhoff, line 85: `/* protected */`?

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,457 @@
...
+    fun
+  }
+
+  def declare(decl: Decl): Unit = {
+    // We convert the declaration into a Z3-level declaration (which is automatically added to Z3's
+    // current state) and record it in out collection(s) of emmitted declarations.
+    // Special handling for macro declarations if expandMacros is true; in that case,
+    // nothing is declared on the Z3 level, and we simply remember the definition ourselves.
+    decl match {
+      case SortDecl(s) => {
```
mschwerhoff, line 351: Braces around case-bodies are not necessary, and IntelliJ suggests to remove them. Please do that.

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -0,0 +1,457 @@
...
+
+      if(Verifier.config.proverEnableResourceBounds) {
+        ctx.updateParamValue("rlimit", (effectiveTimeout * Verifier.config.proverResourcesPerMillisecond).toString)
+      } else {
+        ctx.updateParamValue("timeout", effectiveTimeout.toString)
+      }
+    }
+  }
+
+  override def name: String = Z3ProverAPI.name
```
mschwerhoff, line 446: Could this and the other `def`s here be (`lazy`) `val`s instead?

src/main/scala/rules/Brancher.scala
```diff
@@ -87,40 +95,33 @@ object brancher extends BranchingRules {
...
           SymbExLogger.currentLog().markReachable(uidBranchPoint)
-          executionFlowController.locally(s, v0)((s1, v1) => {
-            if (v.uniqueId != v1.uniqueId) {
+          if (v.uniqueId != v0.uniqueId){
+            /* [BRANCH-PARALLELISATION] */
+            // executing the else branch on a different verifier, need to adapt the state
 
-              /* [BRANCH-PARALLELISATION] */
-              throw new RuntimeException("Branch parallelisation is expected to be deactivated for now")
+            val newFunctions = functionsOfCurrentDecider -- v0.decider.freshFunctions
```
mschwerhoff, line 111: If I remember correctly, the operands are both `ListSet`s, i.e. linked lists, with list subtraction in `O(n^2)`. Not your doing, of course, but maybe you could a comment that points this out.

Side remark: I wanted to replace these sets by more performant ones since forever, but couldn't find a great alternative. Java's `LinkedHashSet` might be a good candidate, but it's not a native Scala collection ...

src/main/scala/rules/Executor.scala
```diff
@@ -100,6 +101,30 @@ object executor extends ExecutionRules {
       Q(s, v)
     } else if (edges.length == 1) {
       follow(s, edges.head, v)(Q)
+    } else if (Verifier.config.parallelizeBranches() &&
```
mschwerhoff, line 104: Wow, that's a lot of (repeated) casts and typechecks. Wouldn't it be more concise to pattern match on the edge types?

src/main/scala/verifier/VerificationPoolManager.scala
```diff
@@ -35,13 +34,16 @@ class VerificationPoolManager(master: MasterVerifier) extends StatefulComponent
...
+    override def emitSettings(contents: Iterable[String]): Unit =
+      slaveVerifiers foreach (_.decider.prover.emitSettings(contents))
   }
 
   /* Verifier pool management */
 
   private def setupSlaveVerifierPool(): Unit = {
     slaveVerifiers = Vector.empty
-    runningVerificationTasks = new ConcurrentHashMap()
+    //runningVerificationTasks = new ConcurrentHashMap()
```
mschwerhoff, line 46: If the code isn't useful as documentation, feel free to remove

mschwerhoff (approved):
I totally approve on branch parallelisation 👍 💯 🥇 🚀 💎 But please see my comments nonetheless.

PR #635, opened 2022-08-22
Removing global state

marcoeilers:
Removing global state in order to be able to run multiple instances of Silicon in parallel (which several frontends would like to do). Depends on Silver PR https://github.com/viperproject/silver/pull/600

We're making the following changes:

- Adapting to Silver changes, in particular, the fact that FastParser is no longer a global object
- Moving Verifier.program, Verifier.functionData, and Verifier.predicateData to the State class, and passing them from there through the code where needed.
- Verifier.inputFile is removed completely. We're explicitly setting it in the MultiRunRecorder and passing it to the runVerifier method; it's not used anywhere else.

Note that Verifier.config remains global, since I'm assuming that we don't want to run Silicon with different settings at the same time, and it's referenced all over the place

### DIFF
--- a/src/main/scala/reporting/MultiRunRecorder.scala
+++ b/src/main/scala/reporting/MultiRunRecorder.scala
@@ -96,6 +96,11 @@ object MultiRunRecorders extends StatefulComponent {
   private val sinks = mutable.ArrayBuffer.empty[PrintWriter]
 
   protected def config: Config = Verifier.config
+
+  /**
+   The source is set by the Silicon verifier (in method [[viper.silicon.Silicon.verify]]) after parsing and type
+   checking but before consistency checks and verification.
+   */
   var source: Option[String] = None
 
   protected def sink(name: String): PrintWriter = {
--- a/src/main/scala/reporting/MultiRunRecorder.scala
+++ b/src/main/scala/reporting/MultiRunRecorder.scala
@@ -96,7 +96,7 @@ object MultiRunRecorders extends StatefulComponent {
   private val sinks = mutable.ArrayBuffer.empty[PrintWriter]
 
   protected def config: Config = Verifier.config
-  protected def source: Option[String] = ??? // Verifier.inputFile.map(_.toString)
+  var source: Option[String] = None
 
   protected def sink(name: String): PrintWriter = {
     val writer =
--- a/src/main/scala/reporting/MultiRunRecorder.scala
+++ b/src/main/scala/reporting/MultiRunRecorder.scala
@@ -96,7 +96,7 @@ object MultiRunRecorders extends StatefulComponent {
   private val sinks = mutable.ArrayBuffer.empty[PrintWriter]
 
   protected def config: Config = Verifier.config
-  protected def source: Option[String] = Verifier.inputFile.map(_.toString)
+  protected def source: Option[String] = ??? // Verifier.inputFile.map(_.toString)
 
   protected def sink(name: String): PrintWriter = {
     val writer =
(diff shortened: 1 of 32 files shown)


src/main/scala/reporting/MultiRunRecorder.scala
```diff
@@ -96,7 +96,7 @@ object MultiRunRecorders extends StatefulComponent {
   private val sinks = mutable.ArrayBuffer.empty[PrintWriter]
 
   protected def config: Config = Verifier.config
-  protected def source: Option[String] = Verifier.inputFile.map(_.toString)
+  var source: Option[String] = None
```
mschwerhoff, line 99: Please add a brief description of when/where the value is set to Some(String). I.e. of the implicit usage protocol for the field.

src/main/scala/verifier/DefaultMasterVerifier.scala
```diff
@@ -262,25 +261,30 @@ class DefaultMasterVerifier(config: Config, override val reporter: Reporter)
...
 
-  private def createInitialState(member: ast.Member, program: ast.Program): State = {
+  private def createInitialState(member: ast.Member, program: ast.Program,
+                                 functionData: Map[ast.Function, FunctionData], predicateData: Map[ast.Predicate, PredicateData]): State = {
     val quantifiedFields = InsertionOrderedSet(ast.utility.QuantifiedPermissions.quantifiedFields(member, program))
     val quantifiedPredicates = InsertionOrderedSet(ast.utility.QuantifiedPermissions.quantifiedPredicates(member, program))
     val quantifiedMagicWands = InsertionOrderedSet(ast.utility.QuantifiedPermissions.quantifiedMagicWands(member, program)).map(MagicWandIdentifier(_, program))
 
-    State(qpFields = quantifiedFields,
+    State(program=program, functionData = functionData,
```
mschwerhoff, line 270: For consistency: please break into two lines, and format as `k = v` (space around `=`)

src/main/scala/verifier/DefaultMasterVerifier.scala
```diff
@@ -262,25 +261,30 @@ class DefaultMasterVerifier(config: Config, override val reporter: Reporter)
...
+          qpFields = quantifiedFields,
           qpPredicates = quantifiedPredicates,
           qpMagicWands = quantifiedMagicWands,
           predicateSnapMap = predSnapGenerator.snapMap,
           predicateFormalVarMap = predSnapGenerator.formalVarMap,
           isMethodVerification = member.isInstanceOf[ast.Member])
   }
 
-  private def createInitialState(@unused cfg: SilverCfg, program: ast.Program): State = {
+  private def createInitialState(@unused cfg: SilverCfg, program: ast.Program,
```
mschwerhoff, line 280: For consistency: if parameters don't fit (> 100 chars) on one line, break over multiple lines, with one parameter per line

src/main/scala/verifier/DefaultMasterVerifier.scala
```diff
@@ -262,25 +261,30 @@ class DefaultMasterVerifier(config: Config, override val reporter: Reporter)
...
+                                 functionData: Map[ast.Function, FunctionData], predicateData: Map[ast.Predicate, PredicateData]): State = {
     val quantifiedFields = InsertionOrderedSet(program.fields)
     val quantifiedPredicates = InsertionOrderedSet(program.predicates)
     val quantifiedMagicWands = InsertionOrderedSet[MagicWandIdentifier]() // TODO: Implement support for quantified magic wands.
 
-    State(qpFields = quantifiedFields,
+    State(program=program, functionData = functionData,
```
mschwerhoff, line 286: See previous comment

mschwerhoff (changes requested):
Thanks for removing the global state! Looks all good to me, modulo the few (and exciting) comments I left.

PR #638, opened 2022-09-15
Bring sequence axiomatization more in line with carbon to fix #629

### LINKED ISSUE #629: Problem with inequality of sequences
Silicon seems to not be able to prove inequality of sequences if the first elements are equal. If I first reason about the elements individually, then the reasoning about the sequences passes as well. This is similar to issue #601, but still seems to be a problem. Like with that issue, Carbon does verify the example.
```
// This passes
method test_seq_unequal()
{
  assert Seq(false, false) != Seq(true, true)
}

// This fails
method test_seq_semiequal()
{
  assert Seq(true, false) != Seq(true, true)
}

// This passes
method test_seq_elemwise()
{
  var s1: Seq[Bool] := Seq(true, true)
  var s2: Seq[Bool] := Seq(true, false)

  assert s1[0] != s2[0] || s1[1] != s2[1]
  assert s1 != s2
}
```

The Silicon version I use is `Silicon 1.1-SNAPSHOT (4c705143)`.

Update: From what we can tell, it seems to be an issue with the axiomatisation of sequence equality, specifically the triggers in the `extensional_seq_equality` axiom. Accessing the unequal item in one of the sequences allows verification to pass:
```
// This passes
method test_seq_semiequal()
{
  var b: Bool
  b := Seq(true, false)[1]
  assert Seq(true, false) != Seq(true, true)
}
```

pieter-bos:
- Weakened triggers for singleton
- Placed append empty left/right in one quantifier
- Split index_over_append into left/right version, and used separate arithmetic functions
- Phrase disequality with skolem function for differing index

### DIFF
--- a/src/main/resources/dafny_axioms/sequences.vpr
+++ b/src/main/resources/dafny_axioms/sequences.vpr
@@ -14,6 +14,9 @@ domain $Seq[E] {
   function Seq_drop(s: $Seq[E], n: Int): $Seq[E]
   function Seq_equal(s1: $Seq[E], s2: $Seq[E]): Bool
   function Seq_sameuntil(s1: $Seq[E], s2: $Seq[E], i: Int): Bool
+  function Seq_skolemdiff(s1: $Seq[E], s2: $Seq[E]): Int
+  function Seq_arith_add(i: Int, j: Int): Int
+  function Seq_arith_sub(i: Int, j: Int): Int
 
   axiom seq_length_non_negative {
     forall s: $Seq[E] :: {Seq_length(s)}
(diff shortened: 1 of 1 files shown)


mschwerhoff:
@marcoeilers Didn't you also recently port Carbon's axioms to Silicon?

marcoeilers:
Yes, here: https://github.com/viperproject/silicon/tree/meilers_seq_set_axioms
That branch uses *exactly* Carbon's sequence, seq, and multiset axioms, in the same way. It's on the to-discuss-list for the next Viper meeting whether that's what we want.

pieter-bos:
I'll close this under the assumption that it's still the plan to work on #642 at some point (though I understand it's difficult and/or not the highest priority). For what it's worth we still run into triggering problems with the default data structures occasionally, though not very often.

Issue #535, opened 2021-03-23
Unexpected output of Z3 causes Silicon to crash


jcp19:
Recent versions of z3 (`4.8.9` and beyond) have introduced changes to the possible outputs of z3 ([issue](https://github.com/Z3Prover/z3/issues/4425), [commit 1](https://github.com/Z3Prover/z3/commit/9f8887cc2ef8248aaeb01b5afe3507923f364b4e), [commit 2](https://github.com/Z3Prover/z3/commit/80d5d661589e6283b5df3a8fd34f2c1eaad7f353)). Silicon does not seem to be prepared to handle them. In particular, Silicon runs fine with [this file](https://github.com/viperproject/silver/files/6188496/scion.gobra.vpr.txt) when I run it with z3 version `4.8.7` but it fails when I use it with version `4.8.9` or higher, in which case it produces the output
```
[info] Verification aborted exceptionally: java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: viper.silicon.reporting.Z3InteractionFailed: Unexpected output of Z3. Expected 'success' but found: (error "line 37114 column 5: canceled")
[info]   Cause: java.util.concurrent.ExecutionException: viper.silicon.reporting.Z3InteractionFailed: Unexpected output of Z3. Expected 'success' but found: (error "line 37114 column 5: canceled")
[info] Silicon found 1 error in 02m:53s:
[info]   [0] Interaction with Z3 (instance 03) failed: Unexpected output of Z3. Expected 'success' but found: (error "line 37114 column 5: canceled")
```

The provided file is large (3157 lines). I was not able to reproduce the error with smaller files.

PS: Thank you @marcoeilers for pointing out the specific changes.

mschwerhoff:
That looks like a consequence of this Z3 issue: https://github.com/Z3Prover/z3/issues/4713. It was closed, but the comment could also be interpreted as "closed and won't fix (for now)".

AFAICS, our only current option is to stick with Z3 4.8.7.

yanntm:
Thanks for the link visible from closed issue Z3Prover/z3#4713. , I guess I'll go with 4.8.7 too, I have the same kind of errors cropping up, it was hard to diagnose too because the sample files needed to reproduce are the bigger ones.

niomaster:
I ran into this while toying with a newer z3. Do you think the proposed workaround (setting a high timeout just for push) is an option for silicon?

mschwerhoff:
[...]

PR #646, opened 2022-10-13
Add option pushTimeout to specify timeout for prover `(push)`

niomaster:
Workaround for #535.

I went with `default = Some(0)` over `None`, meaning `(push)` has no timeout even when the option is not specified, and a finite value is set via `Config.proverTimeout`. (I don't have a strong argument whether it should inherit that value)


mschwerhoff (approved):
👍

marcoeilers (approved):

PR #649, opened 2022-10-21
Field and predicate specific FVFs

marcoeilers:
Making FVF sorts (and the predicate equivalent) field-specific. The point is to avoid triggering the FVF extensionality axioms on pairs of FVFs of different fields (or predicates), which is never useful, and looks potentially expensive. 

Draft, still testing performance implications.

### DIFF
--- a/src/main/resources/field_value_functions_axioms.smt2
+++ b/src/main/resources/field_value_functions_axioms.smt2
@@ -11,7 +11,7 @@
 ; The axiom therefore needs to be emitted after the sort wrappers have
 ; been emitted.
 
-(assert (forall ((vs $FVF<$T$>) (ws $FVF<$T$>)) (!
+(assert (forall ((vs $FVF<$FLD$>) (ws $FVF<$FLD$>)) (!
     (=>
       (and
         (Set_equal ($FVF.domain_$FLD$ vs) ($FVF.domain_$FLD$ ws))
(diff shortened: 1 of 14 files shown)


jcp19:
@marcoeilers does this lead to any noticeable speed-up in our benchmarks?

jcp19:
Did a few extra tests with SCION packages. In one instance (package `slayers/path/onehop`) this lead to a drop in half of the verification time (from 6min16s to 3min36s)

src/main/resources/field_value_functions_axioms.smt2
```diff
@@ -11,24 +11,22 @@
 ; The axiom therefore needs to be emitted after the sort wrappers have
 ; been emitted.
 
-(assert (forall ((vs $FVF<$T$>) (ws $FVF<$T$>)) (!
+(assert (forall ((vs $FVF<$FLD$>) (ws $FVF<$FLD$>)) (!
```
jcp19, line 14: is it guaranteed that any name accepted for a field in Silver can just be replaced in this string such that it still produces valid smtlib2 code?

src/main/scala/supporters/qps/PredicateAndWandSnapFunctionsContributor.scala
```diff
@@ -94,13 +94,12 @@ class DefaultPredicateAndWandSnapFunctionsContributor(preambleReader: PreambleRe
...
-        sorts.PredicateSnapFunction(predicateSnapGenerator.getSnap(predicate)._1))
+    collectedSorts = collectedPredicates.map(predicate =>
+        sorts.PredicateSnapFunction(predicateSnapGenerator.getSnap(predicate)._1, predicate.name)) ++ collectedWandIdentifiers.map(identifier => sorts.PredicateSnapFunction(sorts.Snap, identifier.toString))
 
-    if (collectedPredicates.nonEmpty || collectedWandIdentifiers.nonEmpty) {
-      collectedSorts += sorts.PredicateSnapFunction(sorts.Snap)
-    }
+    //if (collectedPredicates.nonEmpty || collectedWandIdentifiers.nonEmpty) {
+    //  collectedSorts += sorts.PredicateSnapFunction(sorts.Snap, "Wand")
+    //}
```
jcp19, line 103: should we drop this instead of keeping it a comment?

jcp19 (approved):

Issue #376, opened 2019-05-06
Unsoundness: a heap dependent function is instantiated even when its precondition does not hold


viper-admin:
> Created by **@vakaras** on 2019-05-06 11:48
> Last updated on 2019-05-08 09:13

The following example verifies in Silicon while it should not:

```
function Void$discriminant(self: Ref): Int
  requires acc(Void(self))
  ensures false
{
  unfolding acc(Void(self)) in 8
}

predicate Void(self: Ref) {
  false
}

method m_void$$unreachable$opensqu$0$closesqu$() returns (_0: Ref)
{
  assert false
}
```

mschwerhoff:
Following up on the comment from [May 8, 2019](https://github.com/viperproject/silicon/issues/376#issuecomment-588993370): To allow recursive invocations, the dummy trigger would have to be part of the function axiom body as well. Think about whether the well-definedness checks suffice here, i.e. introduce the necessary trigger.

marcoeilers:
Finally fixed.

Issue #369, opened 2019-02-19
Function postconditions are not checked soundly


viper-admin:
> Created by **@alexanderjsummers** on 2019-02-19 14:16
> Last updated on 2019-05-06 12:43

The following example should not verify, but does
```
function test() : Int 
   ensures false
{
  42
}
```

marcoeilers:
Finally fixed.

PR #650, opened 2022-10-22
Sound trigger-independent function encoding

marcoeilers:
This PR solves the unsoundness from issue #376 and #369.
The solution (which Gaurav and I worked out based on earlier discussions by other people, and which Gaurav is in the process of implementing for Carbon) is, for every heap-dependent function with a precondition, to use a new boolean uninterpreted function on the SMT-level that represents whether the function's precondition holds, and make definitional and postcondition axioms conditional on that new function being true. 
Then, we emit the information that a function's precondition holds in different places:
- When Silicon evaluates a function call, after checking that the precondition holds, it assumes that the precondition function holds for the given arguments and in the current state. 
- If function f calls function g, then, after checking the well-definedness of f, we emit an additional axiom that states that, if f's precondition holds, then g's precondition will also hold when f calls g.
- When asserting quantifiers, after checking that they are well-defined, we assert that the asserted body holds under the assumption that for all function applications in its body, their respective preconditions hold.

The PreconditionPropagationTransformer (I'm very open to suggestions for better names and better places to put this code) takes a term that may contain applications of heap-dependent functions, and returns a term stating that the preconditions of all such function applications hold if the conditions under which they are called hold. That is, for a term t, after checking that it is well-defined, one may assume the transformed version of t.

### DIFF
--- a/src/main/scala/rules/Consumer.scala
+++ b/src/main/scala/rules/Consumer.scala
@@ -476,7 +476,7 @@ object consumer extends ConsumptionRules {
       eval(s2, e, pve, v1)((s3, t, v2) => {
         val termToAssert = t match {
           case Quantification(q, vars, body, trgs, name, isGlob) =>
-            val transformed = PreconditionPropagationTransformer.transform(body)
+            val transformed = PreconditionPropagationTransformer.transform(body, s2.program)
             v2.decider.assume(Quantification(q, vars, transformed, trgs, name+"_precondition", isGlob))
             Quantification(q, vars, Implies(transformed, body), trgs, name, isGlob)
           case _ => t
--- a/src/main/scala/rules/Consumer.scala
+++ b/src/main/scala/rules/Consumer.scala
@@ -476,7 +476,7 @@ object consumer extends ConsumptionRules {
       eval(s2, e, pve, v1)((s3, t, v2) => {
         val termToAssert = t match {
           case Quantification(q, vars, body, trgs, name, isGlob) =>
-            val transformed = PreconditionPropagationTransformer.transform(body, s2.program)
+            val transformed = FunctionPreconditionTransformer.transform(body, s3.program)
             v2.decider.assume(Quantification(q, vars, transformed, trgs, name+"_precondition", isGlob))
             Quantification(q, vars, Implies(transformed, body), trgs, name, isGlob)
           case _ => t
--- a/src/main/scala/rules/Consumer.scala
+++ b/src/main/scala/rules/Consumer.scala
@@ -320,16 +320,21 @@ object consumer extends ConsumptionRules {
 
         eval(s, eRcvr, pve, v)((s1, tRcvr, v1) =>
           eval(s1, ePerm, pve, v1)((s2, tPerm, v2) => {
-            val (relevantChunks, _) =
-              quantifiedChunkSupporter.splitHeap[QuantifiedFieldChunk](s2.h, BasicChunkIdentifier(field.name))
-            val (smDef1, smCache1) =
-              quantifiedChunkSupporter.summarisingSnapshotMap(
-                s2, field, Seq(`?r`), relevantChunks, v2)
-            v2.decider.assume(FieldTrigger(field.name, smDef1.sm, tRcvr))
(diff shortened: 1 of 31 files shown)


src/main/scala/rules/Consumer.scala
```diff
@@ -474,7 +474,14 @@ object consumer extends ConsumptionRules {
 
     executionFlowController.tryOrFail0(s1, v)((s2, v1, QS) => {
       eval(s2, e, pve, v1)((s3, t, v2) => {
-        v2.decider.assert(t) {
+        val termToAssert = t match {
+          case Quantification(q, vars, body, trgs, name, isGlob) =>
+            val transformed = PreconditionPropagationTransformer.transform(body, s2.program)
```
mschwerhoff, line 479: `s3.program`?

src/main/scala/state/PreconditionPropagationTransformer.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import viper.silicon.rules.functionSupporter
+import viper.silicon.state.terms.{And, App, HeapDepFun, Implies, Ite, Let, Literal, Not, Or, Quantification, Term, True}
+import viper.silver.ast
+
+object PreconditionPropagationTransformer {
+  def transform(t: Term, p: ast.Program): Term = {
+    val res = t match {
+      case _:Literal => True()
+      case And(ts) => And(transform(ts.head, p), Implies(ts.head, transform(And(ts.tail), p)))
```
mschwerhoff, line 11: AFAICS, this transforms A && B into A' && (A ==> B'), but it's not clear to me, why (and similar for later cases). Would be nice (although somewhat non-traditional) to have a brief description of what the transformer does.

src/main/scala/state/PreconditionPropagationTransformer.scala
```diff
@@ -0,0 +1,36 @@
...
+      case Or(ts) => And(transform(ts.head, p), Implies(Not(ts.head), transform(Or(ts.tail), p)))
+      case Implies(t0, t1) => And(transform(t0, p), Implies(t0, transform(t1, p)))
+      case Ite(t, t1, t2) => And(transform(t, p), Ite(t, transform(t1, p), transform(t2, p)))
+      case Let(bindings, body) => Let(bindings, transform(body, p))
+      case Quantification(q, vars, body, triggers, name, isGlobal) =>
+        val tBody = transform(body, p)
+        if (tBody == True())
+          True()
```
mschwerhoff, line 19: Micro-optimisation, but returning `tBody` would avoid one object creation :-)

src/main/scala/state/PreconditionPropagationTransformer.scala
```diff
@@ -0,0 +1,36 @@
+package viper.silicon.state
+
+import viper.silicon.rules.functionSupporter
+import viper.silicon.state.terms.{And, App, HeapDepFun, Implies, Ite, Let, Literal, Not, Or, Quantification, Term, True}
+import viper.silver.ast
+
+object PreconditionPropagationTransformer {
```
mschwerhoff, line 7: Assuming the transformer propagates preconditions, how about `PreconditionPropagatingTransformer`?

mschwerhoff (approved):
Generally LGTM. I left a few comments, only the one about documentation would entail some work.

src/main/scala/state/PreconditionPropagationTransformer.scala
marcoeilers: Yes. It's because when A&&B is checked for well-definedness, because of short-circuiting, we check that A is well-defined, and if A is true, then we also check if B is well- defined. Here, we're assuming exactly that. Similar argument for other cases. 
I'll add some documentation explaining this.

src/main/scala/state/PreconditionPropagationTransformer.scala
marcoeilers: After writing down what it actually does, I settled on `FunctionPreconditionTransformer`. While it doesn't transform function preconditions, it transforms terms to state that all function preconditions are fulfilled.

PR #651, opened 2022-10-26
Add a flag to disable code emitting heap-dependent trigger expressions

marcoeilers:
This PR adds a flag ``--disableHeapDependentTriggers`` that disables all code in Silicon that emits field and predicate triggers and, in particular, prevents Silicon from emitting snapshot maps etc. that are required only for those triggers. As a result, heap-dependent triggers will no longer work (users can still write them, but the quantifiers that use them will no longer be triggered), i.e., using the flag does not impact soundness, but makes Silicon less expressive (or arguably less complete).

Its purpose is to improve performance for Viper code that does not use such triggers, in particular, Gobra-generated code used in the VerifiedSCION project, which reportedly gets a big performance improvement from disabling them.

### DIFF
--- a/src/main/scala/rules/Consumer.scala
+++ b/src/main/scala/rules/Consumer.scala
@@ -325,7 +325,7 @@ object consumer extends ConsumptionRules {
             val (smDef1, smCache1) =
               quantifiedChunkSupporter.summarisingSnapshotMap(
                 s2, field, Seq(`?r`), relevantChunks, v2)
-            v2.decider.assume(FieldTrigger(field.name, smDef1.sm, tRcvr))
+            //v2.decider.assume(FieldTrigger(field.name, smDef1.sm, tRcvr))
 //            v2.decider.assume(PermAtMost(tPerm, FullPerm()))
             val loss = PermTimes(tPerm, s2.permissionScalingFactor)
             quantifiedChunkSupporter.consumeSingleLocation(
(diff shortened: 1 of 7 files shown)


src/main/scala/rules/Consumer.scala
```diff
@@ -320,16 +320,21 @@ object consumer extends ConsumptionRules {
...
-                s2, field, Seq(`?r`), relevantChunks, v2)
-            v2.decider.assume(FieldTrigger(field.name, smDef1.sm, tRcvr))
-//            v2.decider.assume(PermAtMost(tPerm, FullPerm()))
+            val s2p = if (s2.heapDependentTriggers.contains(field)){
+              val (relevantChunks, _) =
+                quantifiedChunkSupporter.splitHeap[QuantifiedFieldChunk](s2.h, BasicChunkIdentifier(field.name))
+              val (smDef1, smCache1) =
+                quantifiedChunkSupporter.summarisingSnapshotMap(
+                  s2, field, Seq(`?r`), relevantChunks, v2)
+              v2.decider.assume(FieldTrigger(field.name, smDef1.sm, tRcvr))
```
jcp19, line 329: sorry for the noob question but I'm still getting into this part of the codebase. What does it mean to assume a trigger?

jcp19 (approved):
As may be clear from my very noob question, I'm not yet super comfortable with the silicon codebase. Nonetheless, I did not find any obvious mistakes and would be glad to have this as soon as possible for SCION.

src/main/scala/rules/Consumer.scala
marcoeilers: When you write a field trigger { x.f } in Viper, on the SMT level, that'll be translated as triggering on a special function called something like loc_f. That's a boolean function whose value doesn't represent anything and which is just used in triggers. So here we assume that this function is true for some arguments, which adds the function application to the prover state, and will trigger any quantifiers that trigger on that field.

PR #653, opened 2022-11-02
Added flag for disabling automatic function triggering on unfold

marcoeilers:
We trigger function's definitional axioms when unfolding any predicate they depend on.
Even when this functionality is not used, emitting this trigger can cause large slowdowns (since it creates a snapshot, which triggers the snapshot extensionality axiom) in programs that fold and unfold a lot (Vytautas has a test program that takes one minute with the triggers vs. 6 seconds without). 

In the medium term, we should try to minimize the performance impact of these triggers in general, but for now, I'm adding a flag to just disable them.

### DIFF
--- a/src/main/scala/Config.scala
+++ b/src/main/scala/Config.scala
@@ -612,6 +612,12 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
     noshort = true
   )
 
+  val disableFunctionUnfoldTrigger: ScallopOption[Boolean] = opt[Boolean]("disableFunctionUnfoldTrigger",
+    descr = "Disables automatic triggering of function definitions when unfolding predicates they depend on.",
+    default = Some(false),
+    noshort = true
+  )
+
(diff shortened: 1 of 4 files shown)


mschwerhoff:
Good idea! The fewer heap summaries, the better. Instead of an option, why not perform another usage analysis? Because the flag can disable this even if functions depend on predicates, but the additional triggers are not needed for the particular program?

marcoeilers:
I think we should basically *also* do a usage analysis and will do that when I have time.
But with the heap-dependent triggers, clearly, if your program uses one, you want that functionality. Whereas with this, even if you have a (recursive?) function that depends on a predicate, you might prefer to have full control over when your functions are instantiated.

vakaras (approved):
Thanks!

vakaras:
> Instead of an option, why not perform another usage analysis?

If I understand it correctly what this flag does, Prusti programs do not use this feature by construction. So, unless the analysis is fast and reliable, I would prefer having a flag.

PR #657, opened 2022-11-08
Add missing arguments to inverse functions to fix crash

marcoeilers:
In the alternative version of assuming upper bounds for permissions that doesn't use field triggers, I introduced a potential crash because inverse functions aren't given all their arguments. This fixes that.


jcp19 (approved):
LGTM. We can merge it if tests pass

PR #662, opened 2022-11-21
Improvements to --parallelizeBranches

marcoeilers:
Three changes to --parallelizeBranches:
- We parallelize even if an edge is an out-edge
- We saturate when a new verifier takes over a branch
- If parallelizeBranches is used, we use HashSets instead of InsertionOrderedSets for the freshFunctions. For large examples, with InsertionOrderedSets (which are actually ListSets), Silicon spent a LOT of time just computing set unions and differences. Unfortunately, Scala does not have a LinkedHashSet, but --parallelizeBranches leads to massively non-deterministic prover states anyway.


marcoeilers:
Could you send me the example where you got this?

PR #673, opened 2022-12-10
Giving unique default names to all quantifiers so we can identify them in Z3 output

marcoeilers:
Using empty names, which we currently do by default, leads to Z3 using internal names like k!1023 when reporting quantifier instantiation numbers. Those names cannot be traced back to the actual quantifier. By using a meaningless but unique name instead, we can at least search for the reported quantifier IDs in Silicon's SMT logs. 

Long term, obviously, it would be better to try to actually generate meaningful IDs for each emitted quantifier.

### DIFF
--- a/src/main/scala/state/Terms.scala
+++ b/src/main/scala/state/Terms.scala
@@ -462,7 +462,7 @@ sealed trait Quantifier
 
 case object Forall extends Quantifier {
 
-  val qidCounter = new AtomicInteger()
+  private val qidCounter = new AtomicInteger()
 
   def defaultName = s"quant-u-${qidCounter.getAndIncrement()}"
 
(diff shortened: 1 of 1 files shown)


src/main/scala/state/Terms.scala
```diff
@@ -459,8 +461,13 @@ case class False() extends BooleanLiteral {
 sealed trait Quantifier
 
 case object Forall extends Quantifier {
+
+  val qidCounter = new AtomicInteger()
+
+  def defaultName = s"quant-u-${qidCounter.getAndIncrement()}"
```
jcp19, line 467: I would make these private

jcp19 (approved):
LGTM, just a minor comment

PR #674, opened 2022-12-13
Translate quantifier weights

fpoli:
With this PR Silicon translates quantifier weights (added in https://github.com/viperproject/silver/pull/633) to Z3. I added a test to show how to set the weight in the AST.

### DIFF
--- a/src/main/scala/supporters/ExpressionTranslator.scala
+++ b/src/main/scala/supporters/ExpressionTranslator.scala
@@ -10,6 +10,7 @@ import viper.silver.ast
 import viper.silicon.rules.functionSupporter
 import viper.silicon.state.Identifier
 import viper.silicon.state.terms._
+import viper.silver.ast.WeightedQuantifier
 
 trait ExpressionTranslator {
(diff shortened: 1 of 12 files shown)


src/main/scala/supporters/ExpressionTranslator.scala
```diff
@@ -88,10 +89,18 @@ trait ExpressionTranslator {
           }
         )))
 
+        val weight = sourceQuant.info match {
```
marcoeilers, line 92: The same thing needs to happen in the Evaluator: https://github.com/viperproject/silicon/blob/8239c0535fb81b025e1fae39e93df817b7146220/src/main/scala/rules/Evaluator.scala#L631

marcoeilers (commented):
The Evaluator should be adapted similarly to the ExpressionTranslator, apart from that LGTM.

src/main/scala/supporters/ExpressionTranslator.scala
fpoli: Thanks, I pushed a commit. I was looking for `DefaultEvaluator` as in the comment but I couldn't find that file...

marcoeilers (approved):

Issue #641, opened 2022-09-22
Let expression in forall body causes exception


Felalolf:
The following Viper Snipet causes Silicon to crash with "java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected expression acc(y.f, write) cannot be symbolically evaluated". 

```
field f: Int

method bar(x: Seq[Ref]) 
  requires forall i: Int, j: Int :: 0 <= i && i < |x| && 0 <= j && j < |x| && i != j ==> x[i] != x[j]
  requires forall i: Int :: 0 <= i && i < |x| ==> let y == (x[i]) in acc(y.f) // fails
  requires forall i: Int :: 0 <= i && i < |x| ==>  let y == (x[i]) in y.f >= 5 // ok
```

I get the error in the IDE and also when verifying corresponding Viper AST nodes directly.

Felalolf:
The same happens with carbon, see https://github.com/viperproject/carbon/issues/430

Felalolf:
I have attached the stack trace I get when verifying a let expression in a forall body with Gobra:
[log.txt](https://github.com/viperproject/silicon/files/9626309/log.txt)

Issue #99, opened 2014-07-09
Exhaling quantified permissions can add unsound assumptions


viper-admin:
> Created by bitbucket user **sccblom** on 2014-07-09 10:00
> Last updated on 2015-06-11 09:33

The following code seems unsound:

```
#!text

method silly(input: Seq[Ref], gsize: Int)
  requires |input| == gsize
  requires (forall tid: Int :: (tid in [0..gsize)) ==> acc(input[tid].Integer_value, write))
{
  exhale (forall k: Int :: (k in [0..gsize)) ==> acc(input[k].Integer_value, write))
  assert gsize > 0
}
```

PR #677, opened 2023-01-06
Reporting of all errors in the entity result messages

ArquintL:
### Use case for expected errors
Josua Stuck has added the concept of expected verification errors which make Silicon continue despite the occurrence of an error. This feature is in particular used by the refute plugin that marks all encoded refute statements as expected verification errors. These expected errors won't be reported to the user by the refute plugin and instead the plugin produces verification errors for encoded refute statements for which no error has been encountered by Silicon.

### Keeping track of expected errors in Silicon
Silicon keeps track of these expected and encountered verification errors in the `previous` field of Silicon's `VerificationResult`. It seems that the first encountered error is the verification result and all subsequent verification errors (due to continuing with verifying the input) are placed in the `previous` field.

### Post-processing expected errors - so far
The `DefaultMainVerifier` used to produce verification results per member while ignoring the `previous` field, i.e. the first encountered error is returned. [`Silicon` line 247](https://github.com/viperproject/silicon/blob/fb1bd019f764aef130d14a469c3293ab702f970b/src/main/scala/Silicon.scala#L247) retrieved the errors encountered after the first one by taking the `previous` field into account and flatting all verification results.

### Issue
`DefaultMainVerifier` is responsible for emitting entity verification result messages. Since the `DefaultMainVerifier` does not consider the `previous` field of verification results, entity verification result messages contain different (i.e. a subset) of errors compared to Silicon's overall result (when considering the errors for the same member).
This behavior is problematic in the context of [Silver #641](https://github.com/viperproject/silver/pull/641) since plugins will handle entity verification result such that the ViperServer & the IDE support plugins (see [ViperServer #99](https://github.com/viperproject/viperserver/pull/99)).

### Post-processing expected errors - new
This PR basically moves [`Silicon` line 247](https://github.com/viperproject/silicon/blob/fb1bd019f764aef130d14a469c3293ab702f970b/src/main/scala/Silicon.scala#L247) into `DefaultMainVerifier` to ensure that the union of all entity verification results (hopefully) correspond to the overall result returned by Silicon. Note however that grouping, sorting, and dealing with branch conditions have been left in `Silicon` as imho this concerns more errors will be printed and it's not immediately clear whether this functionality can be as easily moved into `DefaultMainVerifier `, even though this could make sense in the long run.

### DIFF
--- a/src/main/scala/Silicon.scala
+++ b/src/main/scala/Silicon.scala
@@ -243,21 +243,21 @@ class Silicon(val reporter: Reporter, private var debugInfo: Seq[(String, Any)] 
 
     /*verifier.bookkeeper.*/elapsedMillis = System.currentTimeMillis() - /*verifier.bookkeeper.*/startTime
 
-    val failures =
-      results.flatMap(r => r :: r.previous.toList)
-        .collect{ case f: Failure => f } /* Ignore successes */
-        .pipe(allResults => {
-          /* If branchconditions are to be reported we collect the different failure contexts
-          *  of all failures that report the same error (but on different branches, with different CounterExample)
-          *  and put those into one failure */
-          if (config.enableBranchconditionReporting())
-            allResults.groupBy(failureFilterAndGroupingCriterion).map{case (_: String, fs:List[Failure]) =>
-              fs.head.message.failureContexts = fs.flatMap(_.message.failureContexts)
-              Failure(fs.head.message)
-            }.toList
-             else allResults.distinctBy(failureFilterAndGroupingCriterion)
-        })
-        .sortBy(failureSortingCriterion)
+    val failures = results
+      .collect{ case f: Failure => f } /* Ignore successes */
+      .pipe(allResults => {
+        /* If branchconditions are to be reported we collect the different failure contexts
+         *  of all failures that report the same error (but on different branches, with different CounterExample)
+         *  and put those into one failure
+         */
+        if (config.enableBranchconditionReporting())
+          allResults.groupBy(failureFilterAndGroupingCriterion).map{case (_: String, fs:List[Failure]) =>
+            fs.head.message.failureContexts = fs.flatMap(_.message.failureContexts)
+            Failure(fs.head.message)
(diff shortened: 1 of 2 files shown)


src/main/scala/Silicon.scala
```diff
@@ -243,21 +243,21 @@ class Silicon(val reporter: Reporter, private var debugInfo: Seq[(String, Any)]
...
-          *  and put those into one failure */
-          if (config.enableBranchconditionReporting())
-            allResults.groupBy(failureFilterAndGroupingCriterion).map{case (_: String, fs:List[Failure]) =>
-              fs.head.message.failureContexts = fs.flatMap(_.message.failureContexts)
-              Failure(fs.head.message)
-            }.toList
-             else allResults.distinctBy(failureFilterAndGroupingCriterion)
-        })
-        .sortBy(failureSortingCriterion)
+    val failures = results
```
mschwerhoff, line 246: Would be good to document (code comment or assert statement) that `r.previous` fields are expected to be irrelevant at this point. Or even empty?

src/main/scala/verifier/DefaultMainVerifier.scala
```diff
@@ -513,4 +517,12 @@ class DefaultMainVerifier(config: Config,
     }
     results
   }
+
+  /**
+    * In case Silicon encounters an expected error (i.e. `ErrorMessage.isExpected`), Silicon continues (until at most
+    * config.numberOfErrorsToReport() have been encountered (per member)).
+    * This function retrieves all errors that have been encountered and creates a verification result containing them.
```
mschwerhoff, line 524: The last sentence doesn't seem fitting:
1. `retrieves all errors` doesn't obviously match a single `::` concatenation
2. `creates a verification result`, but return type is `Seq[VerificationResult]`

mschwerhoff (approved):
I left two comments, otherwise OK.

ArquintL:
bors merge

marcoeilers (approved):

Issue #205, opened 2015-12-28
Different behaviour: impure conditionals vs. conditional permission expressions


viper-admin:
> Created by **@mschwerhoff** on 2015-12-28 20:12
> Last updated on 2016-01-06 17:22

### Regular Permissions ###

Consider the following functions:

```
#!text

function fun01(x: Ref, y: Ref, b: Bool): Int
  requires b ? acc(x.f) : acc(y.f)

function fun02(x: Ref, y: Ref, b: Bool): Int
  requires acc(x.f, b ? write : none)
  requires acc(y.f, !b ? write : none)
```

The footprint/frame/snapshot of `fun01` is tight, i.e. it only consists of `x.f` or `y.f`, depending on `b`. Hence, assigning to, e.g. `y.f` does *not* interfere with `fun01(x, y, true) == old(fun01(x, y, true))`.

In contrast, the footprint/frame/snapshot of `fun02` is *not* tight, i.e. it consists of `x.f` *and*  `y.f`. Hence, assigning to, e.g. `y.f` *does* prevent `fun02(x, y, true) == old(fun02(x, y, true))` from verifying.

### Quantified Permissions ###

A similar situation arises with quantified permissions:

```
#!text

function fun01(xs: Seq[Ref], i: Int): Int
  requires 0 <= i && i < |xs|
  requires forall j: Int :: 0 <= j && j < |xs| && j == i ==> acc(xs[j].f, write)

function fun02(xs: Seq[Ref], i: Int): Int
  requires 0 <= i && i < |xs|
  requires forall j: Int :: 0 <= j && j < |xs| ==> acc(xs[j].f, j == i ? write : none)
```

Analogous to the situation above, assigning to `xs[1].f` does *not* interfere with framing `fun01(xs, 0)`, but it *does* interfere with `fun02(xs, 0)`.

### Comments ###

* Carbon exhibits the same issue (for regular permissions)
* For QPs, it should be possible to overcome the issue by narrowing down the domain definition from the condition *c(x)* to the conditionalised permission expression *c(x) ? p(x) : none*

---

Attachments:
* [**`tighter_fvf_domains.diff`**](https://gist.githubusercontent.com/viper-admin/90ced8d9f48f35a83b5ac45600b32967/raw/60a3e3f9b15d1ca4b982557f6761e7a76601b91b/tighter_fvf_domains.diff)

marcoeilers:
PR #684 fixes the issue for non-quantified permissions.

[...]

Issue #594, opened 2022-02-07
Insufficient Permissions in Function Precondition Not Reported


ArquintL:
The following program verifies with the option `--enableMoreCompleteExhale` but should not (and does not verify if this option is dropped).

Method `client2_pkg_F` only has half a permission to the predicate instance `SomePredicate_pkg_F` and calls function `getter_pkg_F` in its postcondition. However, this function requires write permissions to the predicate instance.

```
field val$_Int: Int

function getter_pkg_F(param_pkg_V0: Ref): Int
  requires acc(SomePredicate_pkg_F(param_pkg_V0), write)
{
  (unfolding acc(SomePredicate_pkg_F(param_pkg_V0), write) in param_pkg_V0.val$_Int)
}

predicate SomePredicate_pkg_F(param_pkg_V0: Ref) {
  true && acc(param_pkg_V0.val$_Int, write)
}

method client2_pkg_F(param_pkg_V0: Ref) returns (res_pkg_V0: Int)
  requires acc(SomePredicate_pkg_F(param_pkg_V0), 1 / 2)
  ensures acc(SomePredicate_pkg_F(param_pkg_V0), 1 / 2)
  ensures res_pkg_V0 == getter_pkg_F(param_pkg_V0)
```

mschwerhoff:
Thanks for reporting this, I'll take a look. Please be aware that `--enableMoreCompleteExhale` is not considered "production-ready", so bugs are somewhat expected.

marcoeilers:
This was fixed in #682.

Issue #523, opened 2020-11-12
Permissions in function precondition not correctly checked when using --enableMoreCompleteExhale


mschwerhoff:
[@niomaster reported](https://github.com/viperproject/silicon/issues/518#issuecomment-718741936) that the following snippet unsoundly verifies when using the *experimental* option `--enableMoreCompleteExhale`:

```
field f: Int

function req(x: Ref): Bool
  requires acc(x.f, 2/1)
{ true }

method test(x: Ref) {
  inhale acc(x.f)
  assert req(x) // Fails only without --enableMoreCompleteExhale (and should fail)
  assert false  // Fails even with --enableMoreCompleteExhale (and should fail)
}
```

marcoeilers:
This was fixed in #682

Issue #557, opened 2021-06-25
List of test cases that fail using more complete exhale


fabianboesiger:
Here's a complete list of all test cases that fail using the `--enableMoreCompleteExhale` option, 14 in total:

```
all/issues/silicon/{0030, 0072, 0232, 0240, 0502, 0505}.vpr
all/predicates/arguments.vpr
quantifiedpermissions/issues/issue_{0149, 0184}.vpr
quantifiedpermissions/misc/misc1.vpr
quantifiedpermissions/sequences/self_framing.vpr
quantifiedpermissions/sets/generalized_shape.vpr
quantifiedpredicates/issues/unfolding.vpr
wands/regression/PackageStateConsolidation.vpr
```

marcoeilers:
An updated list: 
8 tests fail in total.

Most fail because test annotations expect errors caused by incompleteness (i.e., these are not actual test failures):
- all/issues/silicon/0072.vpr
- all/issues/silicon/0232.vpr
- all/predicates/arguments.vpr
- quantifiedpermissions/issues/issue_0184.vpr
- quantifiedpermissions/sequences/self_framing.vpr
- quantifiedpredicates/issues/unfolding.vpr

Actual failures:
- quantifiedpermissions/sequences/mergesort.vpr

Here, the reason seems to be that the accesses to ``a.array`` result in different (newly-defined) values that are defined only inside a quantifier, but are needed to trigger it.

Timeout:
- all/sequences/sequences.vpr 

This test is super flaky in general.

marcoeilers:
Update: quantifiedpermissions/sequences/mergesort.vpr works now. That means that currently, there are no test cases that fail with moreCompleteExhale except for the ones mentioned above that aren't actual test failures.

PR #682, opened 2023-02-13
Sound MoreCompleteExhale on demand

marcoeilers:
This PR does two things:

First, it adds a mode where Silicon's MoreCompleteExhale mode is used on retry only. That is, Silicon works in its normal greedy mode, and whenever it cannot prove an assertion and retries the proof, MoreCompleteExhale is temporarily turned on (for this purpose, a ``moreCompleteExhale`` flag is introduced in the State). Then, the assertion is proved in MCE mode, and the branch continues verifying in greedy mode.
The idea is that MCE is often needed to prove a few assertions in large examples, but is not needed for the vast majority. Since MCE also comes with a performance penalty, we try to do as much as possible in the faster greedy mode, and only switch to MCE when needed.
To do this, the PR adds a new command line flag ``exhaleMode``, which can be set to greedy (default), MCE, or MCE-on-demand. The existing ``enableMoreCompleteExhale`` flag is kept for compatibility reasons but deprecated.

Second, it fixes the unsoundness mentioned in issue #594 and #523. I.e., when checking function preconditions in MCE mode, Silicon used to ignore permission amounts and just prove that *some* permission for every mentioned field and predicate was available. For consistency (until we switch everything to ignoring permission amounts in function preconditions), we now consider permission amounts in these situations as well. The implementation simply uses the existing MCE functionality and, when wildcards are used, also uses the existing ``summarise`` function, which is already used elsewhere and does not have the incompleteness problems of the existing implementation.

### DIFF
--- a/src/main/scala/Config.scala
+++ b/src/main/scala/Config.scala
@@ -635,8 +635,11 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
     case _ => opt
   }
 
+  // DEPRECATED and replaced by exhaleMode.
   val moreCompleteExhale: ScallopOption[Boolean] = opt[Boolean]("enableMoreCompleteExhale",
-    descr = "Enable a more complete exhale version.",
+    descr =  "Warning: This option is deprecated. "
+           + "Please use 'exhaleMode' instead. "
+           + "Enables a more complete exhale version.",
     default = Some(false),
     noshort = true
   )
@@ -795,8 +798,9 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
          (Some(_), Some(_), Some(ExhaleMode.Greedy)) |
          (Some(_), Some(_), Some(ExhaleMode.MoreCompleteOnDemand)) =>
       Left(  s"Option ${counterexample.name} requires setting "
-           + s"flag ${moreCompleteExhale.name}")
-    case (_, Some(true), Some(m)) if m != ExhaleMode.MoreComplete => Left(s"Contradictory values given for options ${moreCompleteExhale.name} and ${exhaleModeOption.name}")
+           + s"${exhaleModeOption.name} to 1 (more complete)")
+    case (_, Some(true), Some(m)) if m != ExhaleMode.MoreComplete =>
+      Left(s"Contradictory values given for options ${moreCompleteExhale.name} and ${exhaleModeOption.name}")
     case _ => Right(())
   }
 
--- a/src/main/scala/Config.scala
+++ b/src/main/scala/Config.scala
@@ -84,10 +84,10 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
 
   private val exhaleModeConverter: ValueConverter[ExhaleMode] = new ValueConverter[ExhaleMode] {
     def parse(s: List[(String, List[String])]): Either[String, Option[ExhaleMode]] = s match {
-      case (_, "0" :: Nil) :: Nil => Right(Some(ExhaleMode.Greedy))
-      case (_, "1" :: Nil) :: Nil => Right(Some(ExhaleMode.MoreComplete))
-      case (_, "2" :: Nil) :: Nil => Right(Some(ExhaleMode.MoreCompleteOnDemand))
-      case Nil => Right(None)
+      case Seq((_, Seq("0"))) => Right(Some(ExhaleMode.Greedy))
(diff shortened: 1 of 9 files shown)


jcp19:
On the "on-demand" mode, does silicon still retry once without mce after a failure, but after performing heap consolidation?

src/main/scala/Config.scala
```diff
@@ -763,18 +793,23 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
       sys.error(s"Unexpected combination: $other")
   }
 
+  validateOpt(counterexample, moreCompleteExhale, exhaleModeOption) {
+    case (Some(_), Some(false), None) |
+         (Some(_), Some(_), Some(ExhaleMode.Greedy)) |
+         (Some(_), Some(_), Some(ExhaleMode.MoreCompleteOnDemand)) =>
+      Left(s"Option ${counterexample.name} requires setting "
+        + s"${exhaleModeOption.name} to 1 (more complete)")
```
jcp19, line 801:
```suggestion
        + s"${exhaleModeOption.name} to 1 (more complete exhale)")
```

src/main/scala/rules/ExecutionFlowController.scala
```diff
@@ -128,9 +129,10 @@ object executionFlowController extends ExecutionFlowRules {
 
         val comLog = new CommentRecord("Retry", s0, v.decider.pcs)
         val sepIdentifier = v.symbExLog.openScope(comLog)
-        action(s0.copy(retrying = true, retryLevel = s.retryLevel), v, (s1, r, v1) => {
+        val temporaryMCE = Verifier.config.exhaleMode != ExhaleMode.Greedy
+        action(s0.copy(retrying = true, retryLevel = s.retryLevel, moreCompleteExhale =  temporaryMCE), v, (s1, r, v1) => {
```
jcp19, line 133:
```suggestion
        action(s0.copy(retrying = true, retryLevel = s.retryLevel, moreCompleteExhale = temporaryMCE), v, (s1, r, v1) => {
```

marcoeilers:
> On the "on-demand" mode, does silicon still retry once without mce after a failure, but after performing heap consolidation? In other words, is this the sequence of steps that silicon performs on a failure?
> 
> 1. perform heap-consolidation and retry
> 2. If failed again, enable mce and retry

On retry, it immediately performs a state consolidation and switches on MCE in one step. I agree it might be worth trying to do only a state consolidation first, but with the current system of nested retries, I don't want to generally add a second retry. If we ever change the retry system in general to avoid the nesting (I still don't know if that's possible or worth it), then this might be a thing we could change.

jcp19 (approved):
LGTM

marcoeilers:
I added a small change that fixes issue #205, which was previously fixed for standard (greedy) Silicon, also in MCE mode.
As a result, in the test suite, only one test actually fails with MCE and another one times out (see the update in issue #557).

jcp19:
> I added a small change that fixes issue #205, which was previously fixed for standard (greedy) Silicon, also in MCE mode.
> 
> As a result, in the test suite, only one test actually fails with MCE and another one times out (see the update in issue #557).

@marcoeilers do you expect this change to have major performance implications when running in mce mode?

marcoeilers:
> > I added a small change that fixes issue #205, which was previously fixed for standard (greedy) Silicon, also in MCE mode.
> > As a result, in the test suite, only one test actually fails with MCE and another one times out (see the update in issue #557).
> 
> @marcoeilers do you expect this change to have major performance implications when running in mce mode?

No. In greedy mode, the exact same fix made basically no difference at all.

src/main/scala/rules/MoreCompleteExhaleSupporter.scala
```diff
@@ -281,35 +281,39 @@ object moreCompleteExhaleSupporter extends SymbolicExecutionRules {
 
         val s0 = s.copy(functionRecorder = currentFunctionRecorder)
 
-        summarise(s0, relevantChunks.toSeq, resource, args, v)((s1, snap, _, _, v1) =>
+        summarise(s0, relevantChunks.toSeq, resource, args, v)((s1, snap, _, _, v1) => {
+          val condSnap = if (v1.decider.check(Greater(perms, NoPerm()), Verifier.config.checkTimeout())) {
+            snap
+          } else {
+            Ite(Greater(perms, NoPerm()), snap.convert(sorts.Snap), Unit)
```
mschwerhoff, line 288: Use one of those optimising term constructors (`IsPositive` etc.) here, instead of `Greater(perms, NoPerm())`?

mschwerhoff (approved):
LGTM

Issue #450, opened 2017-08-06
Report verification results


viper-admin:
>  **Pull request** :twisted_rightwards_arrows: created by **@mschwerhoff** on 2017-08-06 21:34
> Last updated on 2017-08-12 15:45
> Original Bitbucket pull request id: 40
>
> Participants:
>
> * **@aterga** (reviewer)
> * **@mschwerhoff**
>
> Source: unknown commit on branch `default` of an unknown repo
> Destination: https://github.com/viperproject/silicon/commit/f45279de2165878df70d7a1367b85ff6699320f9 on branch `master`
> Marge commit: https://github.com/viperproject/silicon/commit/c9f825ac5938967dc813a4f44c9157a109bdcc59
>
> State: **`MERGED`**

* Made NoopReporter Silicon's default reporter

* Report results of verifying functions/predicates/methods

PR #686, opened 2023-02-17
Renames QuantifierWeightTests to avoid a name clash with Carbon

ArquintL:
Assembling a fat JAR containing our tests currently fails in ViperServer because Carbon and Silicon both have a class QuantifierWeightTests. This PR address this issue by making the test classes' names unique.

See [Carbon #450](https://github.com/viperproject/carbon/pull/450)


jcp19 (approved):
Thanks a lot! LGTM, let's merge it once the tests go through

Issue #695, opened 2023-03-28
Unknown QP unsoundness


JonasAlaif:
The following program verifies in Silicon (latest Nightly, more complete exhale disabled), but fails in Carbon as expected

```
field ref: Ref
field f: Ref
field n: Ref

predicate T(x: Ref)
predicate List(x: Ref) {
  x != null ==> acc(x.ref) && Node(x.ref)
}
predicate Node(x: Ref) {
  acc(x.f) && acc(x.f.ref) && T(x.f.ref) &&
  acc(x.n) && acc(x.n.ref) && List(x.n.ref)
}
define ref_T(x, lft, mut) (valid_lft(lft) && x in lft[tpT] && (mut ==> lft[tpT][x] == 1/1))
define ref_List(x, lft, mut) (valid_lft(lft) && x in lft[tpListT] && (mut ==> lft[tpListT][x] == 1/1))
predicate ListRef(x: Ref, lft: Map[Int, Map[Ref, Perm]], mut: Bool) {
  x != null ==> acc(x.ref) && NodeRef(x.ref, lft, mut)
}
predicate NodeRef(x: Ref, lft: Map[Int, Map[Ref, Perm]], mut: Bool) {
  acc(x.f) && acc(x.f.ref) && ref_T(x.f.ref, lft, mut) &&
  acc(x.n) && acc(x.n.ref) && ListRef(x.n.ref, lft, mut)
}

define tpT 0
define tpListT 1

predicate lft_T(x: Ref) {
  acc(x.ref) && T(x.ref)
}
predicate lft_List(x: Ref) {
  acc(x.ref) && List(x.ref)
}
define Lifetime(lft) (valid_lft(lft) && LifetimeP(lft))
predicate LifetimeP(lft: Map[Int, Map[Ref, Perm]]) {
  valid_lft(lft) &&
  (forall x: Ref :: x in lft[tpT] ==> lft[tpT][x] > 0/1 && lft[tpT][x] <= 1/1 && acc(lft_T(x), lft[tpT][x])) &&
  (forall x: Ref :: x in lft[tpListT] ==> lft[tpListT][x] > 0/1 && lft[tpListT][x] <= 1/1 && acc(lft_List(x), lft[tpListT][x]))
}
function valid_lft(lft: Map[Int, Map[Ref, Perm]]): Bool {
  tpT in lft && tpListT in lft
}

function get_disc(x: Ref, lft: Map[Int, Map[Ref, Perm]]): Ref
[...]

PR #696, opened 2023-03-28
Resetting scaling factor after folding quantified predicate

marcoeilers:
Fixes issue #695: Permission scaling was not reset after folding a predicate that is used in a QP somewhere.


JonasAlaif:
This fixed so many issues with QPs I was having 😍

PR #705, opened 2023-04-14
Fix timeouts and failure generation

marcoeilers:
This PR makes three changes:
- It modifies ``checkSmoke`` to have an option to use the assert timeout (by default, none) instead of the check timeout.
- It modifies smoke checks whose failure would lead to a verification error to use the assert timeout instead (since we don't want verification failures due to timeouts unless the assert timeout is set).
- It moves the code that generates the error message in QP consumes until after the crucial assert, which is important to get the correct counterexample.


jcp19 (approved):
LGTM

PR #706, opened 2023-04-14
Report errors early with --parallelizeBranches

marcoeilers:
When using option --parallelizeBranches, Silicon always verifies all paths in the program, whereas without it, it usually stops after finding an error in the current method. As a result, using --parallelizeBranches can be disadvantageous, since errors are reported only when the whole method with all its paths has been checked.

This PR modifies Silicon such that errors are reported via EntityFailureMessages through the Reporter interface as soon as they are found when using --parallelizeBranches.

### DIFF
--- a/src/main/scala/rules/Brancher.scala
+++ b/src/main/scala/rules/Brancher.scala
@@ -10,10 +10,13 @@ import java.util.concurrent._
 import viper.silicon.common.concurrency._
 import viper.silicon.decider.PathConditionStack
 import viper.silicon.interfaces.{Unreachable, VerificationResult}
+import viper.silicon.reporting.condenseToViperResult
 import viper.silicon.state.State
 import viper.silicon.state.terms.{FunctionDecl, MacroDecl, Not, Term}
 import viper.silicon.verifier.Verifier
 import viper.silver.ast
+import viper.silver.reporter.{EntityFailureMessage, VerificationResultMessage}
+import viper.silver.verifier.Failure
 
 trait BranchingRules extends SymbolicExecutionRules {
   def branch(s: State,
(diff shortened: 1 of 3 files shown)


src/main/scala/state/State.scala
```diff
@@ -64,13 +65,14 @@ final case class State(g: Store = Store(),
...
                        predicateFormalVarMap: Map[ast.Predicate, Seq[terms.Var]] = Map.empty,
-                       isMethodVerification: Boolean = false,
                        retryLevel: Int = 0,
                        /* ast.Field, ast.Predicate, or MagicWandIdentifier */
                        heapDependentTriggers: InsertionOrderedSet[Any] = InsertionOrderedSet.empty,
                        moreCompleteExhale: Boolean = false)
     extends Mergeable[State] {
 
+  val isMethodVerification: Boolean = currentMember.isEmpty || currentMember.get.isInstanceOf[ast.Method]
+
```
jcp19, line 75: why do we have the case `currentMember.isEmpty` here? from the name of the variable, I would expect it to be true only when one can show that the current member is actually a `Method`

src/main/scala/rules/Brancher.scala
```diff
@@ -151,17 +154,25 @@ object brancher extends BranchingRules {
...
-      })
-    } else {
-      Unreachable()
-    }).combine({
+            fThen(v1.stateConsolidator.consolidateIfRetrying(s1, v1), v1)
+          })
+        } else {
+          Unreachable()
+        }
+      if (thenRes.isFatal && !thenRes.isReported && parallelizeElseBranch && s.retryLevel == 0) {
```
jcp19, line 169: Why does the condition have the conjuncts`parallelizeElseBranch`? why not `s.parallelizeBranches` instead? Also, I am reading `s.retryLevel == 0` as meaning "there are no more retries left". Is this correct? Given that this comparison appears often in multiple parts of the code, it could make sense to add an aptly named method to `State` like `isLastRetry(): bool`

jcp19 (approved):
LGTM, just two minor questions

src/main/scala/state/State.scala
marcoeilers: It's None when we're directly verifying a CFG (this functionality is currently unused afaik). I'm pretty sure that for all intents and purposes, that scenario should behave like verifying a method. I'll add a comment.

src/main/scala/rules/Brancher.scala
marcoeilers: Yes, good idea, I'll add isLastRetry.

Regarding the first point: I think that makes no difference at the end of the day. parallelizeElseBranch may be false even when s.parallelizeBranches is true if one of the branches is not feasible. So before adding the isReported field, branching on parallelizeElseBranch would have meant omitting some unnecessary messages; now, there should be no difference. I'll change it since s.parallelizeBranches might be clearer.

Issue #688, opened 2023-03-10
Exhaling an empty QP allows asserting false


vakaras:
The following example:

```viper
domain Snap  {
  function trigger(_0: Snap, _2: Int): Bool
}

function loopkup(address: Ref): Snap
  requires acc(P1(address), wildcard)

predicate P1(address: Ref)

predicate P2(index: Int)

method test()
{
  var address: Ref

//var marco: Perm
//assume marco >= none
//var marcoAddress: Ref
//inhale acc(P1(marcoAddress), marco)
//assume marco <= none

  exhale (forall index: Int ::
    {
        trigger(loopkup(address), index)
    }
    false ==> acc(P2(index), write)
  )

  assert false

}
```

unsoundly verifies with `--carbonQPs --carbonFunctions`. If the commented-out code is uncommented, then it also verifies on regular Silicon.

PR #718, opened 2023-05-15
Fix silver's issue #688

jcp19:
Closes https://github.com/viperproject/silver/issues/688

This PR makes sure that the RHS of a let binding will always reach the SMT solver, even if the corresponding LHS is never used. This is useful when the purpose of a let binding is just to provide triggering terms.


marcoeilers (approved):
LGTM

marcoeilers (approved):

PR #720, opened 2023-05-16
Handling non-existent CEs and fractions correctly in CEs from Z3 via API

marcoeilers:
Both previously crashed Silicon


src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -320,8 +320,13 @@ class Z3ProverAPI(uniqueId: String,
...
   protected def retrieveAndSaveModel(): Unit = {
     if (Verifier.config.counterexample.toOption.isDefined) {
-      val model = prover.getModel
-      lastModel = model
+      try {
+        val model = prover.getModel
+        lastModel = model
+      } catch {
+        case _: Z3Exception =>
+          lastModel = null
```
Aurel300, line 328: Should we log a warning at least?

vakaras (approved):
Yay!

PR #730, opened 2023-06-06
Add experimental --moreJoins option

marcoeilers:
Adds an experimental --moreJoins option that joins paths when possible. This can lead to huge speedups (since it massively reduces the number of paths, at the cost of making them more complex). Should usually be used along with moreCompleteExhale, since greedy Silicon seems to struggle with the resulting more complex states.

This is the code by Fabian Bösiger, merged into current Silicon and adapted to work with --parallelizeBranches.

Depends on Silver PR: https://github.com/viperproject/silver/pull/709


src/main/scala/rules/Executor.scala
```diff
@@ -91,69 +99,113 @@ object executor extends ExecutionRules {
...
+    val jp: Option[SilverBlock] = edges.headOption.flatMap(edge => s.methodCfg.joinPoints.get(edge.source))
+
+    (edges, jp) match {
+      case (Seq(), _) => Q(s, v)
+      case (Seq(edge), _) => follow(s, edge, v, joinPoint)(Q)
+      case (Seq(edge1, edge2), Some(newJoinPoint)) if
+          Verifier.config.moreJoins() &&
+          // Can't directly match type because of type erasure ...
+          edge1.isInstanceOf[ConditionalEdge[ast.Stmt, ast.Exp]] &&
+          edge2.isInstanceOf[ConditionalEdge[ast.Stmt, ast.Exp]] &&
```
jcp19, line 117: Doesn't `isInstanceOf` suffer from all the same problems as match expressions regarding erasure? In other words, why does `isInstanceOf` here work?

I wonder if we would need sth stronger like `TypeTag` to preserve the erased types at runtime (https://www.baeldung.com/scala/type-information-at-runtime)

src/main/scala/rules/Joiner.scala
```diff
@@ -40,16 +48,27 @@ object joiner extends JoiningRules {
...
+            /* In order to prevent mismatches between different final states of the evaluation
+             * paths that are to be joined, we reset certain state properties that may have been
+             * affected by the evaluation - such as the store (by let-bindings) or the heap (by
+             * state consolidations) to their initial values.
+             */
+            s3.copy(g = s1.g,
+                    h = s1.h,
+                    oldHeaps = s1.oldHeaps,
+                    underJoin = s1.underJoin,
+                    // TODO: Evaluation should not affect partiallyConsumedHeap, probably
```
jcp19, line 62: this comment scares me 😅

src/main/scala/state/State.scala
```diff
@@ -219,6 +220,217 @@ object State {
...
+          ssCache2, `hackIssue387DisablePermissionConsumption1`,
+          `qpFields1`, `qpPredicates1`, `qpMagicWands1`, smCache2, pmCache2, smDomainNeeded2,
+          `predicateSnapMap1`, `predicateFormalVarMap1`, `retryLevel`, `useHeapTriggers`, moreCompleteExhale2) =>
+
+            val functionRecorder3 = functionRecorder1.merge(functionRecorder2)
+            val triggerExp3 = triggerExp1 && triggerExp2
+            val possibleTriggers3 = possibleTriggers1 ++ possibleTriggers2
+            val constrainableARPs3 = constrainableARPs1 ++ constrainableARPs2
+
+            val smDomainNeeded3 = smDomainNeeded1 || smDomainNeeded2 // TODO: Use AND or OR to merge?
```
jcp19, line 340: this one scares me as well xD

src/main/scala/state/State.scala
```diff
@@ -219,6 +220,217 @@ object State {
...
+
+            val oldHeaps3 = Map.from(mergeMaps(oldHeaps1, conditions1, oldHeaps2, conditions2)
+            ((_, _) => {
+              None
+            })
+            ((heap1, cond1, heap2, cond2) => {
+              Some(mergeHeap(heap1, cond1, heap2, cond2))
+            }))
+
+            // TODO: InvariantContexts should most likely be the same
```
jcp19, line 384: Are these TODOs still open? If so, maybe we should address them before merging

src/main/scala/rules/Executor.scala
marcoeilers: I think it has that problem in principle, yes, but I'm 99% sure it's not an issue here because we just need to know it's a ``ConditionalEdge``; the type arguments are always the same, so if it's a ``ConditionalEdge``, it's definitely a ``ConditionalEdge[ast.Stmt, ast.Exp]``.

src/main/scala/rules/Brancher.scala
```diff
@@ -23,6 +23,7 @@ trait BranchingRules extends SymbolicExecutionRules {
              condition: Term,
              conditionExp: Option[ast.Exp],
              v: Verifier,
+             sequentialOnly: Boolean = false,
```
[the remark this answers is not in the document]
marcoeilers: Removed it.

Issue #736, opened 2023-07-03
Unsoundness when using let-expressions inside quantifiers.


marcoeilers:
Found by @ArquintL.
The attached program is able to assert false after folding a predicate.

[test.gobra.txt](https://github.com/viperproject/silicon/files/11939964/test.gobra.txt)

PR #737, opened 2023-07-03
Fixing unsound assumption about helper variable inside quantifiers

marcoeilers:
Fixes issue #736.


ArquintL (approved):
Awesome, thanks a lot!

PR #754, opened 2023-10-05
Unfold, fold, and unfolding must have a strictly positive permission amount

marcoeilers:
This addresses https://github.com/viperproject/silver/issues/444 on the Silicon side and depends on a change in Silver to introduce a new kind of error and adapt test annotations.


jcp19 (approved):
LGTM

PR #755, opened 2023-10-06
Update submodules

jcp19:
Currently, I cannot update the submodules of viperserver because silicon and carbon use different versions of silver. This PR addresses that


marcoeilers:
Oh, that's weird, this should be happening automatically. I wonder why it didn't, I hadn't noticed...

Issue #338, opened 2018-05-22
Packaging a wand leads to having an inconsistent state causing unsoundness


viper-admin:
> Created by bitbucket user **abdelzaher** on 2018-05-22 20:51
> Last updated on 2018-05-23 15:09

Here this example verifies as the package statement seems to lead us into some inconsistent state which should not be the case. While computing the footprint of the magic wand half permission is taken from LHS, and half is taken from current state which leads us with fact x.f == 2 && x.f == 3. So we are in an inconsistent state while proving the wand but this should not affect the global state. The two options that we have here are either rejecting the wand or verifying the wand after removing the permissions knowing that LHS can never be achieved.

```
#!scala

field f: Int

predicate p(x:Ref) {acc(x.f)}

method tst(x: Ref)
requires acc(x.f, 1/2) && x.f == 3
{
    package acc(x.f, 1/2) && x.f == 2 --* p(x)
    {
        fold p(x)
        assert x.f == 3
    }

    assert false
}
```

ArquintL:
The example above correctly fails using Viper-IDE release v.4.4.2 (Viper v.24.08-release).
However, the example reported in #410 and the following example are still unsound (tested using the same release; they correctly fail when using Carbon).
```
field x: Int
field y: Int

method unsound(a: Ref)
requires acc(a.x) && acc(a.y)
ensures false
{
    package acc(a.x) --* (acc(a.x) && acc(a.y))

    package acc(a.x) --* (acc(a.x) && acc(a.y)) {
        assert acc(a.x, 2/1)
        assert false // necessary to trigger unsoundness
    }

    assert acc(a.x) // shouldn't hold as we should have zero permissions
    // assert acc(a.x, 2/1) // fails
    apply acc(a.x) --* (acc(a.x) && acc(a.y))
    apply acc(a.x) --* (acc(a.x) && acc(a.y))
}
```

ArquintL:
The example above correctly fails using Viper-IDE release v.4.4.2 (Viper v.24.08-release).
However, the example reported in https://github.com/viperproject/silicon/issues/410 and the following example are still unsound (tested using the same release; they correctly fail when using Carbon).
```
field x: Int
field y: Int

method unsound(a: Ref)
requires acc(a.x) && acc(a.y)
ensures false
{
    package acc(a.x) --* (acc(a.x) && acc(a.y))

    package acc(a.x) --* (acc(a.x) && acc(a.y)) {
        assert acc(a.x, 2/1)
[...]

PR #757, opened 2023-10-10
Separating definitions and other assumptions to fix a wand-related unsoundness

marcoeilers:
This PR extends Silicon's path condition stacks to differentiate between general assumptions and assumptions that represent definitions. For the latter, the new method ``assumeDefinition`` is introduced.
The purpose of this distinction is to fix the unsoundness described in issue #338, which was previously already partly addressed and made much more difficult to trigger by @mschwerhoff.

Now, after packaging a wand, instead of performing a smoke check to see if assumptions from producing a wand left us in an inconsistent state and only remembering path conditions if this is not the case (this was the previous workaround, but it does not prevent us from unsoundly assuming properties other than false), we can omit the smoke check and always assume only the definitions, but no other path conditions.

The previous implementation unsoundly verified the following example:
```
function otherFunc(s: Seq[Int], i: Int): Bool
    requires |s| > 1
{ true }

method foo(x: Ref, y: Ref, s: Seq[Int], i: Int)
{
    package (|s| > 3 ? (i > 1 && i < |s| - 2 && otherFunc(s, i)) : (i > 1 && i < |s| - 2))  --* true {
    }
    //:: ExpectedOutput(assert.failed:assertion.false)
    assert |s| > 0
}
```

### DIFF
--- a/src/main/scala/rules/Evaluator.scala
+++ b/src/main/scala/rules/Evaluator.scala
@@ -171,7 +171,7 @@ object evaluator extends EvaluationRules {
 
       case _: ast.WildcardPerm =>
         val (tVar, tConstraints) = v.decider.freshARP()
-        v.decider.assume(tConstraints)
+        v.decider.assumeDefinition(tConstraints)
         /* TODO: Only record wildcards in State.constrainableARPs that are used in exhale
          *       position. Currently, wildcards used in inhale position (only) may not be removed
          *       from State.constrainableARPs (potentially inefficient, but should be sound).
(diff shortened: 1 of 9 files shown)


marcoeilers:
@mschwerhoff Do you think this makes sense the way I did it?

src/main/scala/rules/Evaluator.scala
```diff
@@ -292,7 +292,7 @@ object evaluator extends EvaluationRules {
       case ast.Let(x, e0, e1) =>
         eval(s, e0, pve, v)((s1, t0, v1) => {
           val t = v1.decider.appliedFresh("letvar", v1.symbolConverter.toSort(x.typ), s1.relevantQuantifiedVariables)
-          v1.decider.assume(t === t0)
+          v1.decider.assumeDefinition(t === t0)
           val newFuncRec = s1.functionRecorder.recordFreshSnapshot(t.applicable.asInstanceOf[Function])
```
mschwerhoff, line 296: Once this PR was accepted, it would be worthwhile to work out if the newly-recorded definitions partly supersedes what the function recorder bookkeeps. If so, the latter could probably be simplified.

marcoeilers: Yes, makes sense. I'll put that on the list. Thanks!

PR #783, opened 2023-12-01
Disabling non-linear integer arithmetics

Dspil:
Adds a flag in silicon to disable non-linear integer arithmetics when Z3 via STDIO is used.

### DIFF
--- a/src/main/scala/Config.scala
+++ b/src/main/scala/Config.scala
@@ -230,6 +230,12 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
     noshort = true
   )
 
+  val disableNLRI: ScallopOption[Boolean] = opt[Boolean]("disableNLRI",
+    descr = "Disable non-linear integer arithetics (prover must be Z3)",
+    default = Some(false),
+    noshort = true
+  )
+
   private val rawProverSaturationTimeout = opt[Int]("proverSaturationTimeout",
     descr = (  "Timeout (in ms) used for the prover's state saturation calls (default: 100). "
              + "A timeout of 0 disables all saturation checks."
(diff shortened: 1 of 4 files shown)


src/main/scala/Config.scala
```diff
@@ -230,6 +230,12 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
     noshort = true
   )
 
+  val disableNLRI: ScallopOption[Boolean] = opt[Boolean]("disableNLRI",
+    descr = "Disable non-linear integer arithetics (prover must be Z3)",
```
jcp19, line 234:
```suggestion
    descr = "Disable non-linear integer arithmetic when using Z3",
```

src/main/scala/Config.scala
```diff
@@ -813,6 +819,11 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
       sys.error(s"Unexpected combination: $other")
   }
 
+  validateOpt(disableNLRI, prover) {
+    case (Some(true), n) if n != Some(Z3ProverStdIO.name) => Left(s"Option ${disableNLRI.name} is only supported with prover ${Z3ProverStdIO.name}")
```
jcp19, line 823:
```suggestion
    case (Some(true), n) if n != Some(Z3ProverStdIO.name) =>
        Left(s"Option ${disableNLRI.name} is only supported with prover ${Z3ProverStdIO.name}")
```

src/main/scala/decider/Z3ProverStdIO.scala
```diff
@@ -65,5 +65,12 @@ class Z3ProverStdIO(uniqueId: String,
     Paths.get(Verifier.config.z3Exe)
   }
 
-  override def emitSettings(contents: Iterable[String]): Unit = emit(contents)
+  override def emitSettings(contents: Iterable[String]): Unit = {
+    emit(contents)
+    if(Verifier.config.disableNLRI.getOrElse(false)) {
+      comment("We disable non-linear integer arithmetics")
```
jcp19, line 71:
```suggestion
      comment("We disable non-linear integer arithmetic")
```

jcp19 (commented):
I have a few minor comments that we should address before merging. Could we also add a test that typically passes, but which should fail with the new flag to make sure that the implementation has the expected behaviour?

src/test/resources/misc/disableNL.vpr
```diff
@@ -0,0 +1,7 @@
+method f(x: Int, y: Int) returns (res: Int)
+  requires 0 <= x <= 10
+  requires 0 <= y <= 10
+  ensures 0 <= res <= 100 // this fails with --disableNL
+{
+  res := x * y
+}
```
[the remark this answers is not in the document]
Dspil: Similar tests about more complete exhale just had a comment there too

marcoeilers:
We already have a flag to pass options to Z3, right? Is this just tomake it more convenient?

jcp19 (approved):
lgtm

PR #784, opened 2023-12-06
Adding an annotation to set prover arguments per method

marcoeilers:
Adding an annotation ``@proverArgs(key=value)`` to set prover configuration parameters per method.

Example usage: 

```
@proverArgs("smt.arith.nl=false")
method m2(i: Int, i2: Int)
    requires i >= 0
    requires i2 >= 0
{
    //:: ExpectedOutput(assert.failed:assertion.false)
    assert i * i2 >= 0
}
```


jcp19 (approved):
Seems good to me

Issue #788, opened 2023-12-20
Slow verification with trivial quantified permissions


bobismijnnaam:
Dear Viper team,

Summary: A common pattern for specifying permissions in VerCors seems to lead to slow verification, even on what seems like simple programs & contracts. Is this an intrinsic limitation, and should we just use not more than 5 quantified permissions per method (that would be pretty bad honestly, but at least now we finally know about it...?), or did I still do something wrong and can I fix it by changing the code/spec? Please have a look at the minified example, flame graph, and comments on the silicon implementation.

Long question: the following program ([triggerTest_small.vpr.txt](https://github.com/viperproject/silicon/files/13726481/triggerTest_small.vpr.txt)) only inhales and exhales quantified permissions. It does not actually mutate state, or shift the ranges of the quantified permissions. All quantified permissions look like this:

```
ensures acc(d.a, write)
ensures d.a != (none1(): Option[Array])
ensures (forall preferred_list_i__: Int ::
    { aloc(optGet1(d.a), preferred_list_i__) }
    0 <= preferred_list_i__ && preferred_list_i__ < alen(optGet1(d.a)) ==>
    acc(aloc(optGet1(d.a), preferred_list_i__).int, write))
```

Repeated 9 times as pre/postconditions, for fields a, b, and c, for variables c, d, and e.

The main program is as follows:

```
method foo1(c1: Ref, d: Ref, e: Ref)
// ... spec ommitted ...
{
  bar1(c1)
  bar1(d)
}
```

`bar1` requires permission for fields a, b and c of the argument,  and permissions of all indices of fields a, b, and c (in the frontend input file, a, b and c are of type `int[]`). Without `bar1(d)`, verification is pretty much instant. With `bar1(d)`, verification takes around 5 minutes. This time almost entirely spent exhaling the quantified permissions in the postcondition of foo. As more postconditions are exhaled, verification time increases, even though these postconditions and the quantifiers look like something viper could easily handle. The last quantified permission takes around 2 minutes, the one before that 74s, etc.

After some investigation (basically printf-style debugging using VerCors' pprof support), it seems these are the places where silicon spends a lot of time. These are all located in QuantifiedChunkSupport.scala:

- In the `precomputedData foreach` loop silicon spends 15 seconds:
  -  7 seconds of this is the chunkDepleted check
  - 7 seconds for the intermediate check at the end of the loop iteration
  - This duration becomes 15 seconds early on, and then seems to stop growing
[...]

PR #789, opened 2023-12-20
Better heuristics for QPs

marcoeilers:
Silicon's heuristics for selecting in what order to use QP chunks when exhaling QPs are somewhat lacking (see issue #788).

This PR adds a much more precise kind of heuristics that should help a lot with performance when QPs of the same "shape" (as in, the same number and types of quantified variables) are inhaled and exhaled repeatedly.
E.g., if the QP we're looking to find or remove has a quantified variable ``i: Int`` and receiver term ``f(a, i)``, and an existing chunk with quantified variable ``x: Int`` has receiver term ``f(g(), x)``, where ``a == g()``, then that chunk would be selected first. We use the decider to check for equality, so it can take all knowledge Silicon currently has into account.
If no such chunk exists, the standard hint based heuristics are used.


jcp19:
For future reference, my quick and dirty measurements in the master branch of SCION (`router` package):
- without this PR: 11:49
- with this PR: ???

marcoeilers:
Haha :D
For the record, I still think this is a good change. It may cause a tiny bit of additional cost for many cases (but should never be really expensive) and helps massively in some cases.

jcp19 (approved):
I'm not an expert in this code, but the changes look ok to me. Unfortunately, SCION did not get faster with this :'(

PR #790, opened 2023-12-21
symbexlog: fix conditionExp for big and/or

pieter-bos:
Hi! Noticed that sometimes the wrong branch condition is reported in the symbolic execution log. The branching record can contain both the expression (silver) and the term (smt-ish) for the branch, we prefer to render the expression when possible.

What I did:
- Checked all usages of `MemberSymbExLogger.insertBranchPoint`; there's just two:
	- In `Executor`, but this doesn't report any condition expressions anyway;
	- In `brancher.branch`, where the `conditionExp` is passed in to the method.

At every usage of `brancher.branch` I checked that the `conditionExp` that is passed is exactly the expression evaluated for the term that is passed. That is: this was the only place it did not match.


marcoeilers:
Thank you!

Issue #387, opened 2019-06-18
Incompleteness in MoreCompleteExhale


viper-admin:
> Created by **@vakaras** on 2019-06-18 16:04
> Last updated on 2019-08-28 10:36

The following example fails with `--enableMoreCompleteExhale`, but works in Carbon and in Silicon without the flag:

```
method test() {

  var _1: Ref
  var _3: Ref
  var _preserve: Ref

  inhale acc(Vec(_preserve), write-read$())
  inhale acc(usize(_3), write)
  inhale acc(_1.val_ref, read$()) && acc(Vec(_1.val_ref), read$()) && acc(usize(_3), read$())
  inhale _preserve == _1.val_ref
  inhale (unfolding acc(usize(_3), write) in _3.val_int) < len(_1.val_ref)

  assert (unfolding acc(usize(_3), write) in _3.val_int) < len(_1.val_ref)
  exhale acc(Vec(_1.val_ref), write - read$())
  assert (unfolding acc(usize(_3), write) in _3.val_int) < len(_1.val_ref)  // Fails with --enableMoreCompleteExhale

  assert  0 <= len(_1.val_ref)
}

field val_int: Int

field val_ref: Ref

function len(_pure_1: Ref): Int
  requires acc(Vec(_pure_1), read$())
  ensures 0 <= result

predicate usize(self: Ref) {
  acc(self.val_int, write)
}

function read$(): Perm
  ensures none < result
  ensures result < write

predicate Vec(self: Ref)
```

mschwerhoff:
@vakaras @alexanderjsummers 

* https://github.com/viperproject/silicon/issues/387#issue-568261787 verifies with the most recent Silicon version, and Z3 4.8.7 x64 on Windows 10
* as does https://github.com/viperproject/silicon/issues/387#issuecomment-588993734
* as does https://github.com/viperproject/silicon/issues/387#issuecomment-588993741
* as does https://github.com/viperproject/silicon/issues/387#issuecomment-588993743
* as does https://github.com/viperproject/silicon/issues/387#issuecomment-588993754
* as does https://github.com/viperproject/silicon/issues/387#issuecomment-588993757

Let me know if you see a reason for keeping the issue open.

vakaras:
I think it can be closed for now. I will reopen in case I find more issues.

JonasAlaif:
@marcoeilers the following example fails with `--enableMoreCompleteExhale`, but works in Carbon and in Silicon without the flag:
```
[...]

PR #795, opened 2024-01-20
Slightly more complete more complete exhale

marcoeilers:
MCE introduces a fresh symbol for the result of a lookup or the snapshot returned by a consume. Sometimes, this causes trouble if the definition of the fresh symbol is, for example, hidden inside a quantifier, and thus the definition is needed to get the quantifier instantiation that contains the definition.

The MCE code (in ``summarise``) already checks if it can syntactically find a chunk that definitely aliases the receiver we're looking for, and in that case, does not introduce a new symbol, but instead returns and additionally constrains the value of that chunk.

Additionally, when consuming, MCE already uses the greedy algorithm (including SMT checks if needed) to check for known aliases. However, if one is found here, it is not used by the summarisation code mentioned above.

This PR makes two changes:
- ``summarise`` gets an additional parameter; if a client has already used the solver to check for known aliases, it can pass along the result here.
- ``summarise`` uses the passed symbol instead of the fresh one in case one is passed.
- If clients of ``summarise`` indicate that they have not used the solver to check for an alias before calling it, *and*  ``summarise`` cannot find an alias syntactically using the existing check, then ``summarise`` itself will use the solver to check for a definite alias, and use its value if one is found.

So an additional prover query is performed only for lookups (not for consumes, which already used the solver) without syntactic aliases, and only if no value is found in the cache.

This fixes an MCE incompleteness that @JonasAlaif mentioned in #387, and also fixes the last remaining MCE incompleteness mentioned in #557 (``quantifiedpermissions/sequences/mergesort.vpr``).

### DIFF
--- a/src/main/scala/rules/MoreCompleteExhaleSupporter.scala
+++ b/src/main/scala/rules/MoreCompleteExhaleSupporter.scala
@@ -32,6 +32,7 @@ object moreCompleteExhaleSupporter extends SymbolicExecutionRules {
                             relevantChunks: Seq[NonQuantifiedChunk],
                             resource: ast.Resource,
                             args: Seq[Term],
+                            knownValue: Option[Option[Term]],
                             v: Verifier)
                            : (State, TaggedSummarisingSnapshot, Seq[Term], Term) = {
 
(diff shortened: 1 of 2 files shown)


mschwerhoff (approved):
LGTM

PR #797, opened 2024-01-22
Using loss with substituted receiver terms in exhaleExt

### LINKED ISSUE #796: Incompleteness when exhaling a QP whose permission expression depends on the quantified variable inside a package statement
@jcp19 found this one.

The following example does not verify but should:
```
field f: Int

method m(xs: Set[Ref], ys: Set[Ref])
{
  inhale forall r: Ref :: r in xs ==> acc(r.f, r in ys ? 1/2 : none)

  package true --* true {
    assert forall r: Ref :: r in xs ==> acc(r.f, r in ys ? 1/2 : none)
  }
}
```

marcoeilers:
…and adding a qvars parameter to enable local checks if permission term is zero.

This fixes #796


jcp19 (approved):
I am not familiar with this part of the code. Nonetheless, nothing wrong pops out to me

PR #807, opened 2024-02-23
Use ordered set to track declared macros, avoid double-declarations

marcoeilers:


src/main/scala/decider/Decider.scala
```diff
@@ -424,19 +424,22 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
...
     }
 
-    def declareAndRecordAsFreshMacros(macros: Seq[MacroDecl], toStack: Boolean): Unit = {
+    def declareAndRecordAsFreshMacros(macros: InsertionOrderedSet[MacroDecl], toStack: Boolean): Unit = {
       if (!toStack) {
-        macros foreach prover.declare
+        for (m <- macros) {
+          if (!_declaredFreshMacros.contains(m))
+            prover.declare(m)
+        }
```
jcp19, line 432: I am bit afraid that using `InsertionOrderedSet` may hurt performance quite a lot. From what I gathered, this data structure is very inneficient and this may hurt us if the lists are very big.

I wonder if we really are making use of the fact that `_declaredFreshMacros` only stores once each value. If that is not necessary, could we keep the implementation using a vector and keep this check? (`(!_declaredFreshMacros.contains(m))`)

marcoeilers: I'm not sure the Vectors will do much better; checking many times if a vector contains a specific value isn't going to be efficient either, right. We just need a good InsertionOrderedSet implementation. But I made a PR for now to just add perform the check.

PR #808, opened 2024-02-23
More efficient tree-like representation for heap-dependent values in function axioms

marcoeilers:


jcp19 (approved):
I am not an expert in this code, but I don't see anything wrong with it

PR #809, opened 2024-02-26
Avoid double-declarations of macros for parallel branch verification

marcoeilers:
Essentially #807 without changing the data type to sets


jcp19 (approved):
LGTM

PR #818, opened 2024-03-14
Find Z3 within PATH env

manud99:
Currently, when using Silicon one has to explicitly define the location of the Z3 binary. Either via the `--z3Exe` option or via the `Z3_EXE` environment variable. My contribution allows the user to have Z3 referenced via its PATH environment variable. This mimics the behavior of the shell. So when it is possible to run `z3 --version` on the command line, one should also be able to run Silicon without any further configuration.

Already set config values are observed before running the PATH lookup, so this PR does not change the behavior of existing setups.

### DIFF
--- a/src/main/scala/Config.scala
+++ b/src/main/scala/Config.scala
@@ -6,7 +6,7 @@
 
 package viper.silicon
 
-import java.nio.file.{Path, Paths}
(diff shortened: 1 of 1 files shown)


marcoeilers:
Thanks!

Issue #307, opened 2017-09-25
Applying expressions potentially unsound


viper-admin:
> Created by **@mschwerhoff** on 2017-09-25 09:03

The current implementation of magic wand snapshots can result in contradicting assumptions (i.e. is unsound) if a magic wand (chunk) is applied more than once. This can be achieved using `applying` expressions, as illustrated in attached file `test205.sil`. A discussion of the problem, as well as a solution, can be found in my thesis and in Nils Becker's project report.

---

Attachments:
* [**`test205.sil`**](https://gist.githubusercontent.com/viper-admin/a11a140bbb306c6f22e41fdffa1e01b7/raw/db4095353c3feb841b6c12c4b6cf6e44d1b3d78a/test205.sil)

PR #836, opened 2024-05-02
Use a Map from Snap to Snap to represent a magic wand snapshot.

manud99:
@JonasAlaif and I are working on a different method to generate snapshots for magic wands. The problem with the current approach is that it is potentially unsound when we apply the same magic wand multiple times. This problem has been illustrated in issue #307 and occurs when using  `applying` expressions.

To solve that issue we introduce wand maps. These are functions that map a snapshot of the wand's LHS to a corresponding snapshot for the RHS. They make sure that values in the state are preserved when applying a magic wand. In comparison to the initial version they also work when applying a magic wand multiple times in different states. One example for that can be found in the file attached in issue #307 or in this example:

```viper
field f: Int

method test08a(x: Ref)
    requires acc(x.f)
{
    package acc(x.f) --* acc(x.f)

    x.f := applying (acc(x.f) --* acc(x.f)) in x.f + 1

    apply acc(x.f) --* acc(x.f)

    assert acc(x.f) && x.f == old(x.f) + 1
    assert false
}
```

The following changes were made:

* Updated the definition of a `MagicWandSnapshot` to contain a variable for a `wandMap` of type `MagicWandSnapFunction`, short MWSF.
* Modified the package wand algorithm such that we create a `wandMap` and add its definition to the path conditions.
* Modified the apply wand algorithm such that we use the `wandMap` when producing the wand's RHS.
*  When there are magic wands in a program we add the definitions and axioms for the functions related to `MagicWandSnapFunction`.

Requires viperproject/silver#788 to fully pass the test suite.

Additionally, there is a test suite consisting of 12 files that check the behavior of this PR in viperproject/silver#796.


src/main/scala/decider/TermToZ3APIConverter.scala
```diff
@@ -435,6 +435,8 @@ class TermToZ3APIConverter
 
       case bop: Combine =>
         ctx.mkApp(combineConstructor, convertTerm(bop.p0), convertTerm(bop.p1))
+      case MagicWandSnapSingleton(abstractLhs, rhsSnapshot) =>
+        ctx.mkApp(combineConstructor, convertTerm(abstractLhs), convertTerm(rhsSnapshot))
```
marcoeilers, line 439: Looks like there are cases missing here?

src/main/scala/rules/MagicWandSupporter.scala
```diff
@@ -250,67 +227,93 @@ object magicWandSupporter extends SymbolicExecutionRules {
...
+      val wandSnapshot: MagicWandSnapshot = if (useMWSF(s4.program)) {
+        val fct = MagicWandSnapFunction(freshSnapRoot, snapRhs, v3.decider.fresh("mwsf", sorts.MagicWandSnapFunction))
+        v3.decider.assumeDefinition(fct.definition)
+        fct
+      } else {
+        MagicWandSnapSingleton(freshSnapRoot, snapRhs)
       }
 
-      val preMark = v3.decider.setPathConditionMark()
+      // If the wand is part of a quantified expression
```
marcoeilers, line 267: No, this checks if this particular want type is used as a quantified resource anywhere in the current function/method, and thus whether chunks for this wand use the internal representation for quantified resources.

src/main/scala/rules/MagicWandSupporter.scala
```diff
@@ -477,4 +530,15 @@ object magicWandSupporter extends SymbolicExecutionRules {
...
+
+  /**
+   * Determine, whether to use MWSF or MagicWandSnapshot.
+   * MWSF have the advantage over MagicWandSnapshot that they can be applied multiple times.
+   * However, they are slower than MagicWandSnapshots.
+   *
+   * @param program AST program which is verified.
+   * @return Boolean indicating whether to use MWSF instead of MagicWandSnapshot.
+   */
+  def useMWSF(program: Program): Boolean =
```
marcoeilers, line 542: Couldn't we decide this on a per-wand basis?

src/main/scala/rules/MagicWandSupporter.scala
```diff
@@ -331,68 +334,118 @@ object magicWandSupporter extends SymbolicExecutionRules {
...
+
+        // If the snapWand is a (wrapped) MagicWandSnapshot then lookup the snapshot of the right-hand side by applying snapLhs.
+        val magicWandSnapshotLookup = snapWand match {
+          case snapshot: MagicWandSnapSingleton =>
+            v2.decider.assume(snapLhs === snapshot.abstractLhs)
+            snapshot.rhsSnapshot
+          case snapshot: MagicWandSnapFunction => snapshot.applyToWandMap(snapLhs)
+          case SortWrapper(snapshot: MagicWandSnapFunction, _) => snapshot.applyToWandMap(snapLhs)
+          // Fallback solution for quantified magic wands
+          case predicateLookup: PredicateLookup =>
```
marcoeilers, line 430: Why could there be a PredicateLookup here? The code below seems to only create MWSSingletons and MWSFunctions.

src/main/scala/state/Terms.scala
```diff
@@ -2295,48 +2300,139 @@ object PredicateTrigger extends PreciseCondFlyweightFactory[(String, Term, Seq[T
...
+   * Apply the given snapshot of the left-hand side to the magic wand map to get the snapshot of the right-hand side
+   * which includes the values of the left-hand side.
+   *
+   * @param snapLhs The snapshot of the left-hand side that should be applied to the magic wand map.
+   * @return The snapshot of the right-hand side that preserves the values of the left-hand side.
+   */
+  def applyToWandMap(snapLhs: Term): Term = MWSFLookup(wandMap, snapLhs)
+}
+
+object MagicWandSnapFunction  {
```
marcoeilers, line 2410: Same for this, shouldn't this extend the Flyweight factory?

src/main/scala/state/Terms.scala
```diff
@@ -2295,48 +2300,139 @@ object PredicateTrigger extends PreciseCondFlyweightFactory[(String, Term, Seq[T
...
+  override def sort: Sort = sorts.Snap
+
+  override def p0: Term = abstractLhs
+
+  override def p1: Term = rhsSnapshot
 }
 
-object MagicWandSnapshot  {
+object MagicWandSnapSingleton {
+  var pool = new TrieMap[(Term, Term), MagicWandSnapSingleton]()
```
marcoeilers, line 2334: Is there a reason you have to declare the pool yourself here and can't inherit it from ConditionalFlyweightFactory?

src/main/scala/supporters/BuiltinDomainsContributor.scala
```diff
@@ -110,9 +112,9 @@ abstract class BuiltinDomainsContributor extends PreambleContributor[Sort, Domai
     })
   }
 
-  protected def collectFunctions(domains: Set[ast.Domain], program: ast.Program): Unit = {
-    domains foreach (
-      _.functions foreach (df =>
+  protected def collectFunctions(domains: Set[(ast.DomainType, ast.Domain)], program: ast.Program): Unit = {
```
marcoeilers, line 115: This change doesn't seem to accomplish anything?

src/main/scala/supporters/DefaultMapsContributor.scala
```diff
@@ -8,8 +8,10 @@ package viper.silicon.supporters
 
 import scala.reflect.{ClassTag, classTag}
 import viper.silicon.Config
+import viper.silicon.common.collections.immutable.InsertionOrderedSet
```
marcoeilers, line 11: Since nothing else changed in the file I'd assume that the added imports aren't needed

src/main/scala/supporters/DefaultSetsContributor.scala
```diff
@@ -12,6 +12,7 @@ import viper.silver.ast
 import viper.silicon.common.collections.immutable.InsertionOrderedSet
 import viper.silicon.state.terms.{Sort, Term, sorts}
 import viper.silicon.verifier.Verifier
+import viper.silicon.utils.ast.ViperEmbedding
```
marcoeilers, line 15: Same here

src/main/scala/verifier/DefaultMainVerifier.scala
```diff
@@ -6,6 +6,7 @@
 
 package viper.silicon.verifier
 
+import viper.silicion.supporters.MagicWandSnapFunctionsContributor
```
marcoeilers, line 9: We already have ``import viper.silicon.supporters._``

marcoeilers (changes requested):

src/main/scala/supporters/BuiltinDomainsContributor.scala
manud99: When I was trying to add the definitions for `Map<$Snap~_$Snap>` to the SMT2 preamble I was working on that class and modified it. For this method in particular I removed the following line and instead modified this function here.

```scala
val sourceDomainInstantiations = sourceDomainInstantiationsWithType.map(x => x._2)
```

The changes in that class are purely cosmetic and not necessary for my PR. Should I revert all changes in this file?

src/main/scala/rules/MagicWandSupporter.scala
manud99: If the wand is used as a quantified resource, we get a `PredicateLookup` when consuming that magic wand. An example for this is the following program that I've extracted from the test file [QPWands.vpr](https://github.com/viperproject/silver/blob/8e84e064d890fc28ed20678a9740ccc5760d2d91/src/test/resources/wands/new_syntax/QPWands.vpr).

```viper
field f: Int

method test14(xs: Seq[Ref], y: Ref)
    requires forall x: Ref :: x in xs ==> acc(x.f)
    requires |xs| >= 1
    requires acc(y.f)
{
    xs[0].f := 0
    var some: Ref := xs[0]

    package acc(y.f) --* acc(some.f) && acc(y.f)

    // Makes a quantified magic wand out of it
    var completed: Seq[Ref] := Seq(some)
    assert forall x: Ref :: x in completed ==> acc(y.f) --* acc(x.f) && acc(y.f)

    assert applying (acc(y.f) --* acc(some.f) && acc(y.f)) in some.f == 0
}
```

Without the two lines after the comment, we have a normal magic wand and we get a `MagicWandSnapshot` here. With those two lines we have `PredicateLookup` here. Because I did not find out how to handle that case with a `MagicWandSnapFunction` I chose to use a fallback that does the same as the initial version.

manud99:
@marcoeilers Thanks for your code review. I updated my branch accordingly.

After performing some benchmarking, Jonas and I decided to revert the second optimization. We compared the mean execution time of all test files that contain a magic wand. In average this version took 2.2% longer than the initial version. The second optimization took slightly longer (+2.72%).

src/main/scala/rules/MagicWandSupporter.scala
marcoeilers: Right, the PR only changes the encoding of non-quantified wands, makes sense.
I guess that means for quantified wands, we still always have the same encoding that's potentially unsound when it's applied multiple times, right? Is that intentional or is it an accidental omission?

src/main/scala/supporters/BuiltinDomainsContributor.scala
marcoeilers: Ah, I see. No that seems fine then.

src/main/scala/rules/MagicWandSupporter.scala
manud99: This was intentional. I've tried to do it, but after one day of debugging a 2000-5000 line SMT2 file I decided to move on to other more interesting test cases for now. And yes, the current version with `PredicateLookup` is potentially unsound. I've written a test case that illustrates this behavior [here](https://github.com/manud99/silver/blob/e0c11acdefdbbd9dd4abac445dfe041f2033d7c4/src/test/resources/wands/snap_functions/test12.vpr). It verifies unsound using both versions of Silicon - the current release and the one from this PR.

src/main/scala/state/Terms.scala
```diff
@@ -2295,48 +2300,70 @@ object PredicateTrigger extends PreciseCondFlyweightFactory[(String, Term, Seq[T
...
+ * Represents a snapshot of a magic wand, which is a map from Snap to Snap.
+ *
+ * @param abstractLhs The term that represent the snapshot of the consumed left-hand side of the magic wand.
+ *                    It is used as a bound variable in a forall quantifier.
+ * @param rhsSnapshot The term that integrates the left-hand side in the snapshot that is produced after applying the magic wand.
+ * @param wandMap The map that represents the snapshot of the magic wand. It is a variable of sort Map(Snap, Snap).
+ *                In the symbolic execution this represents a function that when we apply a magic wand
+ *                it consumes the left-hand side and uses the resulting snapshot to look up which right-hand side should be produced.
+ */
+class MagicWandSnapshot(val abstractLhs: Var, val rhsSnapshot: Term, val wandMap: Term) extends Term with ConditionalFlyweight[(Var, Term, Term), MagicWandSnapshot] {
```
JonasAlaif, line 2313: One thing I'm not sure about is why a wand snapshot carries around `val abstractLhs: Var, val rhsSnapshot: Term`: aren't these just a construct of packaging a wand and shouldn't exist outside of that? Is there something else that we need these two fields for?

manud99: Good question. I once used those fields as a compact representation for the definition of the MWSF. But the final solution does not use that anymore. Therefore, I removed those fields.

JonasAlaif:
@manud99 could you update the PR from the main branch so that I can merge it (don't have access to a pc right now)

manud99:
@JonasAlaif Done. I updated this branch and the corresponding branch in silver.

PR #853, opened 2024-06-13
Not pushing conditions using permission introspection further in

marcoeilers:
The ``--conditionalizePermissions`` command line option rewrites implications and ternary expressions to avoid branching. 
The general pattern is that ``e1 ==> acc(e2.f, e3)`` is rewritten to ``acc(e2.f, e1 ? e3 : none``. 

This rewriting is not always sound if the condition (``e1`` in this case) uses permission introspection.
Example:
```
inhale P1(x)
exhale perm(P1(x)) >= write ==> (P1(x) && P1(x))
```
must not be rewritten to
```
inhale P1(x)
exhale acc(P1(x), perm(P1(x)) >= write ? write : none) && acc(P1(x), perm(P1(x)) >= write ? write : none)
```
since the value of ``perm(P1(x)) >= write`` changes.

This PR conservatively prevents any rewriting where the condition contains ``perm`` or ``forperm``.
This fixes one issue mentioned in #851.

### DIFF
--- a/src/test/resources/conditionalizePermissions/let.vpr
+++ b/src/test/resources/conditionalizePermissions/let.vpr
@@ -1,0 +1,15 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+field f: Int
+
+function ok(): Bool
+
+
+function val(): Ref
+    requires ok()
+
+method test()
+{
+    inhale ok() ==> let x == (val()) in acc(x.f)
+}
--- a/src/test/resources/conditionalizePermissions/let.vpr
+++ b/src/test/resources/conditionalizePermissions/let.vpr
@@ -11,5 +11,8 @@ function val(): Ref
 
 method test()
 {
+    // Test that the condition is not pushed inside the let expression to get e.g.
+    // inhale let x == (val()) in acc(x.f, ok() ? write : none)
+    // which would not be well-defined.
     inhale ok() ==> let x == (val()) in acc(x.f)
 }
--- a/src/test/resources/conditionalizePermissions/silicon_0008.vpr
+++ b/src/test/resources/conditionalizePermissions/silicon_0008.vpr
@@ -1,0 +1,15 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+
+field f: Int
+
+method deAlias(x: Ref, y: Ref, k: Perm)
+  requires k >= none
+  requires acc(x.f, k) && acc(y.f, k)
+  // The following postcondition should fail in default greedy Silicon because --conditionalizePermissions creates a
+  // situation where the two chunks may or may not alias in the same branch, so they cannot definitively be merged,
+  // so greedy Silicon cannot prove the postcondition using any individual chunk.
+  //:: ExpectedOutput(postcondition.violated:insufficient.permission)
+  ensures x == y ==> acc(x.f, 2 * k)
+{}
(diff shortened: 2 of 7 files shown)


src/test/scala/SiliconTestsConditionalizePermissions.scala
```diff
@@ -0,0 +1,15 @@
...
+
+package viper.silicon.tests
+
+class SiliconTestsConditionalizePermissions extends SiliconTests {
+  override val testDirectories: Seq[String] = Seq("conditionalizePermissions")
+
+ override val commandLineArguments: Seq[String] = Seq(
+    "--timeout", "300" /* seconds */,
+    "--conditionalizePermissions")
+}
```
jcp19, line 15: Nice!

src/test/resources/conditionalizePermissions/silicon_0008.vpr
```diff
@@ -0,0 +1,15 @@
...
+field f: Int
+
+method deAlias(x: Ref, y: Ref, k: Perm)
+  requires k >= none
+  requires acc(x.f, k) && acc(y.f, k)
+  // The following postcondition should fail in default greedy Silicon because --conditionalizePermissions creates a
+  // situation where the two chunks may or may not alias in the same branch, so they cannot definitively be merged,
+  // so greedy Silicon cannot prove the postcondition using any individual chunk.
+  //:: ExpectedOutput(postcondition.violated:insufficient.permission)
+  ensures x == y ==> acc(x.f, 2 * k)
```
jcp19, line 14: doesn't this postcondition fail regardless of the algorithm because we do not constrain `k` to be at most `1/2`?

src/test/resources/conditionalizePermissions/let.vpr
```diff
@@ -0,0 +1,15 @@
...
+function ok(): Bool
+
+
+function val(): Ref
+    requires ok()
+
+method test()
+{
+    inhale ok() ==> let x == (val()) in acc(x.f)
+}
```
jcp19, line 15: is this test checking that we can push the perm amount to an access predicate "inside" a let expression? If so, would there be a way to adapt this test to fail if that is not the case?

jcp19 (approved):
The code looks ok to me, I have a few questions about the tests though

src/test/resources/conditionalizePermissions/silicon_0008.vpr
marcoeilers: I think I don't understand what you mean. The postcondition promises 2k permissions, right, so no matter what k is (and there isn't anything that would require it at least 1/2 as far as I can see), if x == y, that should hold and we have 2k permission (but Silicon greedy can't prove it after the rewriting).

src/test/resources/conditionalizePermissions/let.vpr
marcoeilers: This test is supposed to ensure that we *don't* push the ``ok()`` condition too far in. If we push it into the let-expression, i.e., to get ``inhale let x == (val()) in acc(x.f, ok() ? write : none)``, then ``val()`` isn't well-defined, and the test would fail.,

src/test/resources/conditionalizePermissions/silicon_0008.vpr
jcp19: Ah, nevermind, Somehow I failed to realize that we were using a `k` in the perm amount of `x.f` and `y.f`. Please ignore this comment

src/test/resources/conditionalizePermissions/let.vpr
jcp19: Can you please add this as a comment to the test, please?

jcp19 (approved):

src/test/resources/conditionalizePermissions/let.vpr
marcoeilers: Yes, that makes sense, done.

PR #879, opened 2024-11-26
Optional Snapshot Generation in Consume

mlimbeck:
During the execution of certain statements, consume generates snapshots that are unused. For consuming QP-Chunks, this results in unnecessary snapshot map definitions being triggered, increasing the number of quantifier instantiations and negatively impacting performance.

To address this, snapshot generation is now prevented for these cases. While the primary performance improvement applies to consuming QP-Chunks, the change is also applied to consuming non-QP-Chunks to ensure consistent behavior across all scenarios.

**Example:** 
```
var rs: Set[Ref]
var x: Ref
var val: Int
assume x in rs
inhale forall r: Ref :: r in rs ==> acc(r.f)

val := x.f // This will trigger all snapshot map definitions related to x.f.

exhale forall r: Ref :: r in rs ==> acc(r.f, 1/3) 
exhale forall r: Ref :: r in rs ==> acc(r.f, 1/3) 
// The exhales will create new snapshot map definitions which are never needed but still triggered.
```

**Affected Statements:**

Field assignments
Method calls (consume of pres)
Exhale statements
Assert statements
While loops (consume of invariants)

### DIFF
--- a/src/main/scala/rules/ChunkSupporter.scala
+++ b/src/main/scala/rules/ChunkSupporter.scala
@@ -29,10 +29,11 @@ trait ChunkSupportRules extends SymbolicExecutionRules {
               argsExp: Option[Seq[ast.Exp]],
               perms: Term,
               permsExp: Option[ast.Exp],
+              SnapNeeded: Boolean,
               ve: VerificationError,
               v: Verifier,
               description: String)
-             (Q: (State, Heap, Term, Verifier) => VerificationResult)
+             (Q: (State, Heap, Option[Term], Verifier) => VerificationResult)
              : VerificationResult
 
   def produce(s: State, h: Heap, ch: NonQuantifiedChunk, v: Verifier)
@@ -71,16 +72,17 @@ object chunkSupporter extends ChunkSupportRules {
               argsExp: Option[Seq[ast.Exp]],
               perms: Term,
               permsExp: Option[ast.Exp],
+              SnapNeeded: Boolean,
               ve: VerificationError,
               v: Verifier,
               description: String)
-             (Q: (State, Heap, Term, Verifier) => VerificationResult)
+             (Q: (State, Heap, Option[Term], Verifier) => VerificationResult)
              : VerificationResult = {
 
-    consume2(s, h, resource, args, argsExp, perms, permsExp, ve, v)((s2, h2, optSnap, v2) =>
+    consume2(s, h, resource, args, argsExp, perms, permsExp, SnapNeeded, ve, v)((s2, h2, optSnap, v2) =>
       optSnap match {
         case Some(snap) =>
-          Q(s2, h2, snap.convert(sorts.Snap), v2)
+          Q(s2, h2, Some(snap.convert(sorts.Snap)), v2)
         case None =>
           /* Not having consumed anything could mean that we are in an infeasible
            * branch, or that the permission amount to consume was zero.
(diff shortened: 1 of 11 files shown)


jcp19:
I wonder, did you benchmark this on sth like scion or sth generated by Prusti? I wonder if there are any implications in performance (and potentially completeness)

marcoeilers:
Completeness should not be impacted in any way, the snapshots are only omitted when they definitely won't be used.

src/main/scala/rules/Consumer.scala
```diff
@@ -110,34 +114,41 @@ object consumer extends ConsumptionRules {
...
-        wrappedConsumeTlc(s, h, a, pve, v)((s1, h1, snap1, v1) => {
-          consumeTlcs(s1, h1, tlcs.tail, pves.tail, v1)((s2, h2, snap2, v2) =>
-            Q(s2, h2, Combine(snap1, snap2), v2))})
+        wrappedConsumeTlc(s, h, a, returnSnap, pve, v)((s1, h1, snap1, v1) => {
+          consumeTlcs(s1, h1, tlcs.tail, returnSnap, pves.tail, v1)((s2, h2, snap2, v2) =>
+
+            (snap1, snap2) match {
+              case (Some(sn1), Some(sn2)) => Q(s2, h2, Some(Combine(sn1, sn2)), v2)
+              case (None, None) => Q(s2, h2, None, v2)
+              case (_, _) =>  sys.error(s"Consume returned unexpected snapshot: ${(returnSnap, (snap1, snap2))}")
```
marcoeilers, line 138: I guess if you wanted to be more certain that everything always does the right thing, you could match on Some, Some if returnSnap, and on None, None if !returnSnap.

src/main/scala/rules/ChunkSupporter.scala
```diff
@@ -111,24 +115,28 @@ object chunkSupporter extends ChunkSupportRules {
     if (s.exhaleExt) {
       val failure = createFailure(ve, v, s, "chunk consume in package")
       magicWandSupporter.transfer(s, perms, permsExp, failure, Seq(), v)(consumeGreedy(_, _, id, args, _, _, _))((s1, optCh, v1) =>
-        Q(s1, h, optCh.flatMap(ch => Some(ch.snap)), v1))
+        if(returnSnap){
```
marcoeilers, line 118: Here and elsewhere: I think Silicon code always puts a space here: ``if (...``.

src/main/scala/rules/MoreCompleteExhaleSupporter.scala
```diff
@@ -343,22 +354,27 @@ object moreCompleteExhaleSupporter extends SymbolicExecutionRules {
 
         val s0 = s.copy(functionRecorder = currentFunctionRecorder)
 
-        summarise(s0, relevantChunks.toSeq, resource, args, argsExp, Some(definiteAlias.map(_.snap)), v)((s1, snap, _, _, _, v1) => {
-          val condSnap = if (v1.decider.check(IsPositive(perms), Verifier.config.checkTimeout())) {
-            snap
+        summarise(s0, relevantChunks.toSeq, resource, args, argsExp, Some(definiteAlias.map(_.snap)), returnSnap, v)((s1, snap, _, _, _, v1) => {
```
marcoeilers, line 357: Can you maybe avoid the call to summarise completely if returnSnap is false?

marcoeilers (approved):
Thanks!
I marked three things, everything else looks fine to me.

src/main/scala/rules/MoreCompleteExhaleSupporter.scala
mlimbeck: Yes, it does require a bit of code duplication, but not much. I also made a few other changes in `MoreCompleteExhaleSupporter`:

- **`permSummariseOnly()`**: Computes the permission sum.
- **`summariseOnly()`**: Now calls `permSummariseOnly()` instead of computing the permission sum.
- **`summarise()`**: Only invoked when a snapshot is needed, so the `returnSnap` argument is no longer required.

Additionally, the `ssCache` can now hold entries both with and without snapshot definitions. This avoids redundant computation of permission expressions and improves efficiency.

If you have time could you take another look at the changes in `MoreCompleteExhaleSupporter`?

PR #904, opened 2025-03-04
Fixing issue #903

### LINKED ISSUE #903: Unsound recursive function encoding with wildcard permissions
Having a recursive, terminating function that takes wildcard permissions to the same heap location for all recursive calls seems to trigger an unsoundness.

```
// Any copyright is dedicated to the Public Domain.
// http://creativecommons.org/publicdomain/zero/1.0/

field f: Int

predicate Mem(x: Ref)
{
    acc(x.f) && 0 <= x.f
}

function recurse(x: Ref, idx: Int): Int
requires acc(Mem(x), wildcard)
requires unfolding acc(Mem(x), wildcard) in 0 <= idx && idx <= x.f
decreases unfolding acc(Mem(x), wildcard) in x.f - idx
{
    unfolding acc(Mem(x), wildcard) in idx == x.f ? 42 : recurse(x, idx + 1)
}

method fooUnsound(x: Ref)
requires Mem(x)
ensures  Mem(x)
ensures  false // succeeds unexpectedly
{
    var oldResult: Int := recurse(x, 0)
    unfold Mem(x)
    x.f := x.f + 1
    fold Mem(x)
    var newResult: Int := recurse(x, 0) // this call triggers the unsoundness
}
```

Note that replacing all `wildcard` permission amounts by `1/1` does _not_ result in the unsoundness, i.e., the postcondition fails as expected. Strangely, the type checker warns about `1/1` being treated like `wildcard`, which is however clearly not the case!

Here is a slightly longer example using sequences but involving the same unsoundness:
[unsound.vpr.zip](https://github.com/user-attachments/files/19056320/unsound.vpr.zip)

marcoeilers:
The underlying issue was that fresh values introduced by state consolidation during function verification were made global variables that were then constrained by every function invocation, and different invocations could lead to a contradiction. 
Instead, these global variables should be functions that depend on the function arguments, and we should constrain their values for the current function arguments (similar to another recent issue https://github.com/viperproject/silicon/pull/895).

Closes #903

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: monthly
+      day: monday
+    groups:
+      all:
+        patterns:
+          - "*"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
 
       # Checkout Silicon (note: all checkouts delete the contents of their working directory)
       - name: Checkout Silicon
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           submodules: true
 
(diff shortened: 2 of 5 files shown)


ArquintL (approved):

superaxander:
@marcoeilers I noticed that this PR changed the `StateConsolidator#combineSnapshots` function which I (partially) duplicated in #860 as `QuantifiedChunkSupport#combineFieldSnapshotMaps` and `QuantifiedChunkSupport#combinePredicateSnapshotMaps`. If I'm understanding this issue correctly it would also occur there?

viper-admin:
Great catch! Yes, we'll have to adapt that one as well.

PR #908, opened 2025-03-10
Match on `PIdnUseExp` instead of `PIdnUse`

LaurenzV:
Necessary for https://github.com/viperproject/silver/pull/820.


LaurenzV:
I'm not sure what's going on, but now it fails even earlier when checking out the branch: https://github.com/viperproject/silver/actions/runs/13896191787/job/38877256432?pr=820 😅 I'm not sure why, but I presume my merge commit from main somehow trips it up when trying to check out silver now... I presume the PR should have been squashed before merging, any idea how to fix?

LaurenzV:
I think not, but I'm also not sure what exactly the problem is, because the submodule still seems to link to viperproject/silver... Is it somehow possible to manually trigger a update submodules from Silver Bot?

PR #916, opened 2025-04-22
Making assumption of function preconditions for injectivity check local

marcoeilers:
…and fixing a resulting incompleteness.

This is an attempt to do what @superaxander did in PR #914 but 
- without the push/pop calls, since we never really manually call them virtually anywhere in Silicon; I think just asserting the implication should achieve the same effect
- also fixing an incompleteness that he found that can appear in this context (a receiver term that was needed to trigger a QP did not always make it to the SMT level) 

@superaxander: This version works for the example you included; could you check if it results in any other incompletenesses?

### DIFF
--- a/src/main/scala/rules/Evaluator.scala
+++ b/src/main/scala/rules/Evaluator.scala
@@ -260,19 +260,17 @@ object evaluator extends EvaluationRules {
                   if (s1.triggerExp) {
                     (True, Option.when(withExp)(TrueLit()()), s1)
                   } else {
-                    val s1a = tRcvr match {
-                      case _: Literal | _: Var => s1
+                    val (s1a, lhs) = tRcvr match {
+                      case _: Literal | _: Var => (s1, True)
                       case _ =>
                         // Make sure the receiver exists on the SMT level and is thus able to trigger any relevant quantifiers.
                         val rcvrVar = v1.decider.appliedFresh("rcvr", tRcvr.sort, s1.relevantQuantifiedVariables.map(_._1))
-                        val debugExp = Option.when(withExp)(DebugExp.createInstance("temp var for field access receiver"))
-                        v1.decider.assumeDefinition(BuiltinEquals(rcvrVar, tRcvr), debugExp)
                         val newFuncRec = s1.functionRecorder.recordFreshSnapshot(rcvrVar.applicable.asInstanceOf[Function])
-                        s1.copy(functionRecorder = newFuncRec)
+                        (s1.copy(functionRecorder = newFuncRec), BuiltinEquals(rcvrVar, tRcvr))
                     }
                     val permVal = relevantChunks.head.perm
                     val totalPermissions = permVal.replace(relevantChunks.head.quantifiedVars, Seq(tRcvr))
-                    (IsPositive(totalPermissions), Option.when(withExp)(ast.PermGtCmp(ast.CurrentPerm(fa)(fa.pos, fa.info, fa.errT), ast.NoPerm()())(fa.pos, fa.info, fa.errT)), s1a)
+                    (Implies(lhs, IsPositive(totalPermissions)), Option.when(withExp)(ast.PermGtCmp(ast.CurrentPerm(fa)(fa.pos, fa.info, fa.errT), ast.NoPerm()())(fa.pos, fa.info, fa.errT)), s1a)
                   }
                 v1.decider.assert(permCheck) {
                   case false =>
--- a/src/main/scala/rules/Evaluator.scala
+++ b/src/main/scala/rules/Evaluator.scala
@@ -256,22 +256,28 @@ object evaluator extends EvaluationRules {
                   val triggerExp = Option.when(withExp)(DebugExp.createInstance(s"FieldTrigger(${eRcvr.toString()}.${fa.field.name})"))
                   v1.decider.assume(trigger, triggerExp)
                 }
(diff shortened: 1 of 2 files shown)


marcoeilers:
(Or maybe I should check first why an example in the test suite no longer terminates.)

superaxander:
I just ran the VerCors test suite with these changes and the examples that started failing with #914 (which prompted me to find the minimal example I shared) did not fail now.

superaxander:
Well I haven't been able to minimise this any further, removing any part of this such as for example the call to `setChooseFresh2` results in a passing verification. However without any changes silicon will complain that ` scale1(blockdimZ * write) * scale1(blockdimY * write)` may be negative. I also tested this with and without `--numberOfParallelVerifiers 1` since for the version in #914 this made a difference but with the version here the behaviour is the same with or without.

<details><summary>Code</summary>
<p>

```
domain Block  {

  function block_length(b: Block): Int

  function loc(b: Block, i: Int): Ref

  function loc_inv_1(r: Ref): Block

  function loc_inv_2(r: Ref): Int

  axiom {
    (forall b: Block :: { block_length(b) } block_length(b) >= 0)
  }

  axiom {
    (forall b: Block, i: Int ::
      { loc(b, i), loc(b, i) }
      loc_inv_1(loc(b, i)) == b && loc_inv_2(loc(b, i)) == i)
  }
}

domain Pointer  {

  function pointer_block(p: Pointer): Block

  function pointer_offset(p: Pointer): Int

  function pointer_inv(r: Ref): Pointer

  axiom {
    (forall p: Pointer ::
      { pointer_offset(p), pointer_offset(p) }
      pointer_offset(p) >= 0 &&
      pointer_offset(p) < block_length(pointer_block(p)))
  }

  axiom {
    (forall r: Ref :: { pointer_inv(r) } ptrDeref(pointer_inv(r)) == r)
  }

  axiom {
    (forall p: Pointer :: { ptrDeref(p) } pointer_inv(ptrDeref(p)) == p)
  }

}

field int: Int


function ptrDeref(p: Pointer): Ref
{
  loc(pointer_block(p), pointer_offset(p))
}

function ptrAdd(p: Pointer, offset: Int): Pointer
  requires 0 <= pointer_offset(p) + offset
  requires pointer_offset(p) + offset < block_length(pointer_block(p))
  ensures offset == 0 ==> result == p
  ensures pointer_offset(result) == pointer_offset(p) + offset
  ensures pointer_block(result) == pointer_block(p)


function scale1(amount: Perm): Perm
  requires amount >= none
  ensures result >= none
{
  amount
}

function setRange(l: Int, r: Int): Set[Int]
  ensures (forall i: Int ::
      { (i in result) }
      (i in result) == (l <= i && i < r))
  ensures (l <= r ? |result| == r - l : |result| == 0)


method setChooseFresh2(xs: Set[Int]) returns (res: Int)
  requires !(|xs| == 0)
  ensures (res in xs)


method example1(tid: Int, blockdimY: Int, blockdimZ: Int, griddimX: Int,
  griddimY: Int, griddimZ: Int, a: Pointer)
  requires 0 < blockdimY
  requires 0 < blockdimZ
  requires 0 < griddimX
  requires 0 < griddimY
  requires 0 < griddimZ
  requires 0 <= pointer_offset(a)
  requires pointer_offset(a) < block_length(pointer_block(a))
  requires (forall threadidxx: Int ::
      { ptrDeref(ptrAdd(a, threadidxx)) }
      0 <= threadidxx && threadidxx < 1 ==>
      acc(ptrDeref(ptrAdd(a, threadidxx)).int,
      scale1(griddimZ * write) *
      scale1(griddimY * write) *
      scale1(griddimX * write) *
      scale1(blockdimZ * write) *
      scale1(blockdimY * write) *
      write))
{
    var indet: Bool
    if (indet) {
        exhale (forall threadidxx: Int ::
            { ptrDeref(ptrAdd(a, threadidxx)) }
            0 <= threadidxx && threadidxx < 1 ==>
            acc(ptrDeref(ptrAdd(a, threadidxx)).int,
            scale1(griddimZ * write) *
            scale1(griddimY * write) *
            scale1(griddimX * write) *
            scale1(blockdimZ * write) *
            scale1(blockdimY * write)))
    } else {
        var dummy: Int := setChooseFresh2(setRange(0, griddimZ))
        while (true)
            invariant 0 < blockdimY && 0 < blockdimZ &&
                ((0 <= pointer_offset(a)) &&
                (pointer_offset(a) < block_length(pointer_block(a))) &&
                (forall threadidxx: Int ::
                { ptrDeref(ptrAdd(a, threadidxx)) }
                0 <= threadidxx && threadidxx < 1 ==>
                acc(ptrDeref(ptrAdd(a, threadidxx)).int,
                scale1(blockdimZ * write) *
                scale1(blockdimY * write))))
        {
            inhale false
        }
        inhale false
    }
    inhale false
}
```

</p>
</details>

superaxander:
Ok nevermind, the issue I'm encountering is unrelated to this PR (it might actually be an incompleteness but it's really hard to diagnose)

marcoeilers:
Okay, thanks a lot!
If this version doesn't end up causing any issues, I'll merge it, since it's definitely unnecessary to keep the assumption around, and this is basically the clean way to fix the problem you found. If there are other issues I'd have another look at the other alternatives you tried out.

superaxander:
Yeah I have not been able to find any issues (the only issue I found ended up being reproducible even without these changes)

PR #933, opened 2025-06-25
Add implementation of setTimeout for CVC5

superaxander:
I'm not sure if there is a specific reason why timeouts are not set for the CVC5 prover? The comment says that cvc5 does not support setting the timeout after instantiation but in my testing it works fine. (although it can sometimes exceed the timeout because it's not very strict about keeping to it) I looked through the release notes but couldn't find mention of this being a new feature.


marcoeilers:
Thanks a lot! I have no idea why whoever wrote the code thought this wasn't supported, but it's clearly working now.

PR #936, opened 2025-07-11
Stricter triggers on FVF snap extensionality axiom 

marcoeilers:
This should prevent some unnecessary expensive instantiations of that axiom. In particular, this change lets us avoid instantiations for FVFs that don't actually have a domain, in which case it'll never be proved equal to another one anyway.

### DIFF
--- a/src/main/scala/state/Terms.scala
+++ b/src/main/scala/state/Terms.scala
@@ -2198,6 +2198,17 @@ object Domain extends CondFlyweightTermFactory[(String, Term), Domain] {
   override def actualCreate(args: (String, Term)): Domain = new Domain(args._1, args._2)
 }
 
+class HasDomain(val field: String, val fvf: Term) extends Term with ConditionalFlyweight[(String, Term), HasDomain] {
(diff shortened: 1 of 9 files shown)


src/main/scala/state/Terms.scala
```diff
@@ -2198,6 +2198,17 @@ object Domain extends CondFlyweightTermFactory[(String, Term), Domain] {
...
+class HasDomain(val field: String, val fvf: Term) extends Term with ConditionalFlyweight[(String, Term), HasDomain] {
+  utils.assertSort(fvf, "field value function", "FieldValueFunction", _.isInstanceOf[sorts.FieldValueFunction])
+
+  val sort = sorts.Bool
+  override val equalityDefiningMembers: (String, Term) = (field, fvf)
+}
+
+object HasDomain extends CondFlyweightTermFactory[(String, Term), HasDomain] {
+  override def actualCreate(args: (String, Term)): HasDomain = new HasDomain(args._1, args._2)
+}
```
jcp19, line 2210: @marcoeilers can you please add, as a comment, an explanation of what this term means? (Your recent message to me would be a possible candidate)

jcp19 (approved):
I'm not an expert in this part of the code, but from what I see, it looks fine.

PR #939, opened 2025-08-27
Fixing solver crash with branch parallelization

marcoeilers:


jcp19:
what caused this crash? May it be that this is the reason why we were consistently having silicon crashes when verifying the SCION router with `--parallelizeBranches`?

PR #948, opened 2025-11-12
Avoid defining let-bound variables on SMT level

marcoeilers:
When evaluating an expression ``let x == (e1) in e2``, Silicon emits an ``(assert (= x e1))`` for a fresh variable ``x`` on the SMT level to make sure that ``e1`` is available as a triggering term even when does not occur in ``e2``; this is done to fix https://github.com/viperproject/silver/issues/688.

However, these additional variables and equalities can slow down verification significantly (see attached file 
[floats.txt](https://github.com/user-attachments/files/23506821/floats.txt)).

This PR implements a compromise and emits these definitions only when ``e1`` might be triggering and only during function and predicate verification, since in those, let-bound function calls and similar expressions are commonly used e.g. to call lemmas or trigger quantifiers (as in the original issue).

### DIFF
--- a/src/main/scala/rules/Evaluator.scala
+++ b/src/main/scala/rules/Evaluator.scala
@@ -250,10 +250,18 @@ object evaluator extends EvaluationRules {
 
       case l@ast.Let(x, e0, e1) =>
         eval(s, e0, pve, v)((s1, t0, e0New, v1) => {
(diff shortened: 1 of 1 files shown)


JonasAlaif:
What if we detected that a let bound variable is never used and added it to the SMT query iff that's the case.

PR #966, opened 2026-04-11
Record metadata for each SMT-query

jcp19:
**Architecturally, the way we record the per-query data might not be the most adequate; perhaps the symbex logger would be a better place for that.**

This PR introduces support for recording and classifying SMT interactions (assertions, checks, and scope push/pop operations) issued during verification. Each recorded interaction carries two orthogonal labels:

**`kind` (`QueryKind`)** — the type of solver interaction:
  - `Assert` — from `Decider#assert`
  - `Check` — from `Decider#check` and `Decider#checkSmoke`
  - `Push` / `Pop` — from `Decider#pushScope` / `Decider#popScope`

**`category` (`ProofQueryKind`)** — the semantic purpose of the query:
  - `Consistency` — well-formedness obligations (injectivity of quantified-permission receivers, ...)
  - `Heap` — heap-access and permission-amount correctness (chunk existence, alias, sufficient permission, non-negativity/positivity of permission amounts, ...)
  - `FunctionalCorrectness` — user-visible proof obligations (pre/postconditions, `assert` statements, sequence-index bounds, non-zero divisor, ...)
  - `PathInfeasibility` — branch feasibility and smoke checks
  - `ScopeManagement` — push/pop of the prover assertion stack

Each recorded interaction also carries:
- **`member`** — name of the program member (method/function/predicate/domain) whose verification triggered the query, if known;
- **`pos`** — source position of the proof obligation;
- **`durationMs`** — wall-clock time of the prover call;
- **`succeeded`** — whether the query returned true / Unsat;
- **`description`** — an optional short text description populated at call sites for extra clarity

The feature is enabled by the CLI option `--recordProofQueries <file>`, which writes a CSV to the given path at the end of verification.

## Implementation Details
- Recording is only performed when `Verifier.config.recordProofQueries` is defined; there is no runtime cost when the flag is off.
- Queries that are trivially entailed by the current path conditions (short-circuited by `isKnownToBeTrue`) are **not** recorded — only queries that actually reach the prover.
- The `kind` parameter is **required** on `Decider#check` and `Decider#assert`, so every call site must state its intent explicitly. This makes classification decisions easy to audit in review, and it makes the compiler reject any new query site that is added without a classification.
- `checkSmoke` defaults `kind` to `PathInfeasibility` but accepts an override for smoke checks that are semantically used for a different purpose (e.g. verifying that permission was actually taken from the heap in `MagicWandSupporter`).
- `member` and `pos` are threaded through `findChunk` / `findChunkWithProver` and the chunk-order heuristics so nearly every recorded query is attributed to a specific program member and source position.
- CSV export renders missing metadata as `?`.

## Known Limitation
`member` is `None` only on the CFG-based verification path (`DefaultMainVerifier#createInitialState(cfg, ...)`), which is used by external CFG-driven callers rather than by terminal Viper runs. A bare `SilverCfg` carries no owning `ast.Member`, so there is nothing to thread through `cfgSupporter.verify`; queries issued on that path are exported with `member = ?`.

### DIFF
--- a/src/main/scala/interfaces/decider/ProofQueryKind.scala
+++ b/src/main/scala/interfaces/decider/ProofQueryKind.scala
@@ -1,0 +1,32 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2019 ETH Zurich.
+
+package viper.silicon.interfaces.decider
+
+/** Categorises the purpose of an SMT query (assert or check) issued by the verifier. */
+sealed trait ProofQueryKind
+
+object ProofQueryKind {
+  /** (a) Consistency checks: non-negative/positive permission assertions, injectivity checks
+   *  in quantified permissions, and similar well-formedness obligations. */
+  case object Consistency extends ProofQueryKind
+
+  /** (b) Heap proof obligations: chunk-existence checks, permission-amount checks during
+   *  consume/produce operations, and related heap-access correctness queries. */
+  case object Heap extends ProofQueryKind
+
+  /** (c) Functional-correctness queries: pre/postcondition checks, assert-statement assertions,
+   *  array-index bounds, divisor non-zero, and similar user-visible proof obligations. */
+  case object FunctionalCorrectness extends ProofQueryKind
+
+  /** (d) Axiomatisation queries: consistency of function, domain, or predicate axioms
+   *  (rare – axioms are mostly assumed, not asserted). */
+  case object Axiomatization extends ProofQueryKind
+
+  /** (e) Path-infeasibility checks: smoke checks and branch-feasibility tests that determine
+   *  whether the current execution path is reachable at all. */
+  case object PathInfeasibility extends ProofQueryKind
+}
--- a/src/main/scala/interfaces/decider/ProofQueryKind.scala
+++ b/src/main/scala/interfaces/decider/ProofQueryKind.scala
@@ -29,4 +29,8 @@ object ProofQueryKind {
   /** (e) Path-infeasibility checks: smoke checks and branch-feasibility tests that determine
    *  whether the current execution path is reachable at all. */
   case object PathInfeasibility extends ProofQueryKind
+
+  /** (f) Unknown: used when the purpose of the query does not clearly fall into any of the
+   *  above categories, or has not yet been classified. */
+  case object Unknown extends ProofQueryKind
 }
--- a/src/main/scala/interfaces/decider/ProofQueryKind.scala
+++ b/src/main/scala/interfaces/decider/ProofQueryKind.scala
@@ -33,4 +33,8 @@ object ProofQueryKind {
   /** (f) Unknown: used when the purpose of the query does not clearly fall into any of the
    *  above categories, or has not yet been classified. */
   case object Unknown extends ProofQueryKind
+
+  /** (g) Scope-management operations: push and pop of the prover assertion stack
+   *  used to bound the scope of branch assumptions, contract checks, etc. */
+  case object ScopeManagement extends ProofQueryKind
 }
--- a/src/main/scala/interfaces/decider/ProofQueryKind.scala
+++ b/src/main/scala/interfaces/decider/ProofQueryKind.scala
@@ -10,31 +10,24 @@ package viper.silicon.interfaces.decider
 sealed trait ProofQueryKind
 
 object ProofQueryKind {
-  /** (a) Consistency checks: non-negative/positive permission assertions, injectivity checks
-   *  in quantified permissions, and similar well-formedness obligations. */
+  /** (a) Consistency checks: injectivity of quantified-permission receivers and similar
+   *  well-formedness obligations. */
   case object Consistency extends ProofQueryKind
 
-  /** (b) Heap proof obligations: chunk-existence checks, permission-amount checks during
-   *  consume/produce operations, and related heap-access correctness queries. */
+  /** (b) Heap proof obligations: chunk-existence checks, permission-amount checks (including
+   *  non-negativity and positivity of permission expressions) during consume/produce
+   *  operations, and related heap-access correctness queries. */
   case object Heap extends ProofQueryKind
 
   /** (c) Functional-correctness queries: pre/postcondition checks, assert-statement assertions,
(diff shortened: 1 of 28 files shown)


src/main/scala/resources/NonQuantifiedPropertyInterpreter.scala
```diff
@@ -120,7 +122,9 @@ class NonQuantifiedPropertyInterpreter(heap: Iterable[Chunk], verifier: Verifier
                                     otherwise: PropertyExpression[K],
                                     info: Info): (Term, Option[ast.Exp]) = {
     val conditionTerm = buildPathCondition(condition, info)._1
-    if (verifier.decider.check(conditionTerm, Verifier.config.checkTimeout())) {
+    if (verifier.decider.check(conditionTerm, Verifier.config.checkTimeout(),
+                               kind = ProofQueryKind.Heap, member = member,
+                               description = Some("chunk property condition"))) {
```
jcp19: TODO: improve description

src/main/scala/rules/MagicWandSupporter.scala
```diff
@@ -196,7 +197,8 @@ object magicWandSupporter extends SymbolicExecutionRules {
              * from heap, i.e. that tEq does not result in already having the required permissions before
              * consuming from heap.
              */
-            if (v.decider.checkSmoke()) {
+            if (v.decider.checkSmoke(member = sOut.currentMember.map(_.name),
+                                     description = Some("smoke check: magic wand"))) {
```
jcp19: kind missing here

src/main/scala/rules/MoreCompleteExhaleSupporter.scala
```diff
@@ -390,15 +416,21 @@ object moreCompleteExhaleSupporter extends SymbolicExecutionRules {
...
               Ite(IsPositive(perms), snap.convert(sorts.Snap), Unit)
             })
           if (!moreNeeded) {
             Q(s1, newHeap, condSnap, v1)
           } else {
-            v1.decider.assert(pNeeded === NoPerm) {
+            v1.decider.assert(pNeeded === NoPerm,
+                              kind = ProofQueryKind.Heap, pos = resource.pos,
+                              member = s1.currentMember.map(_.name),
+                              description = Some("MCE sufficient permission (split)")) {
```
jcp19: unclear description

src/main/scala/rules/QuantifiedChunkSupport.scala
```diff
@@ -919,7 +920,10 @@ object quantifiedChunkSupporter extends QuantifiedChunkSupport {
     val nonNegImplicationExp = eCond.map(c => ast.Implies(c, ast.PermGeCmp(ePerm.get, ast.NoPerm()())())(c.pos, c.info, c.errT))
     val nonNegTerm = Forall(qvars, Implies(FunctionPreconditionTransformer.transform(nonNegImplication, s.program), nonNegImplication), Nil)
     // TODO: Replace by QP-analogue of permissionSupporter.assertNotNegative
-    v.decider.assert(nonNegTerm) {
+    v.decider.assert(nonNegTerm,
+                     kind = ProofQueryKind.Consistency, pos = forall.pos,
```
jcp19: Probably should be of type `Heap`

src/main/scala/rules/QuantifiedChunkSupport.scala
```diff
@@ -1158,11 +1165,14 @@ object quantifiedChunkSupporter extends QuantifiedChunkSupport {
     val nonNegTerm = Forall(qvars, Implies(FunctionPreconditionTransformer.transform(nonNegImplication, s.program), nonNegImplication), Nil)
     val nonNegExp = qvarExps.map(qv => ast.Forall(qv, Nil, nonNegImplicationExp.get)())
     // TODO: Replace by QP-analogue of permissionSupporter.assertNotNegative
-    v.decider.assert(nonNegTerm) {
+    v.decider.assert(nonNegTerm,
+                     kind = ProofQueryKind.Consistency, pos = resource.pos,
```
jcp19: should be `Heap`

src/main/scala/interfaces/decider/ProofQueryKind.scala
```diff
@@ -0,0 +1,36 @@
...
+   *  consume/produce operations, and related heap-access correctness queries. */
+  case object Heap extends ProofQueryKind
+
+  /** (c) Functional-correctness queries: pre/postcondition checks, assert-statement assertions,
+   *  array-index bounds, divisor non-zero, and similar user-visible proof obligations. */
+  case object FunctionalCorrectness extends ProofQueryKind
+
+  /** (d) Axiomatisation queries: consistency of function, domain, or predicate axioms
+   *  (rare – axioms are mostly assumed, not asserted). */
+  case object Axiomatization extends ProofQueryKind
```
jcp19: Is this used at all?

src/main/scala/interfaces/decider/ProofQueryKind.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  /** (e) Path-infeasibility checks: smoke checks and branch-feasibility tests that determine
+   *  whether the current execution path is reachable at all. */
+  case object PathInfeasibility extends ProofQueryKind
+
+  /** (f) Unknown: used when the purpose of the query does not clearly fall into any of the
+   *  above categories, or has not yet been classified. */
+  case object Unknown extends ProofQueryKind
```
jcp19: Is this used at all?

src/main/scala/rules/QuantifiedChunkSupport.scala
```diff
@@ -2122,7 +2151,9 @@ object quantifiedChunkSupporter extends QuantifiedChunkSupport {
             receiverTerms.zip(cInvertibles).forall(p => {
               if (cQvars.length == quantVars.length && cQvars.zip(quantVars).forall(vars => vars._1.sort == vars._2.sort)) {
                 val secondReplaced = p._2.replace(cQvars, quantVars)
-                v.decider.check(SimplifyingForall(quantVars, p._1 === secondReplaced, Seq()), Verifier.config.checkTimeout())
+                v.decider.check(SimplifyingForall(quantVars, p._1 === secondReplaced, Seq()), Verifier.config.checkTimeout(),
+                  kind = ProofQueryKind.Heap, member = member, pos = pos,
+                  description = Some("QP receiver forall match"))
```
jcp19: make these function calls more readable, don't put multiple long arguments in the same line. apply this everywhere

PR #969, opened 2026-04-21
Debugging improvements

marcoeilers:
- Two changes to make sure we keep old (debug) heaps recorded in the state during expression evaluation, which were previously discarded because expression evaluation was not expected to generate additional old heaps
- Changing debug expressions for function applications, unfoldings, and applying expressions from a generic ``joined_unfolding``/... variable to the actual evaluated expression wrapped into an old


marcoeilers:
@jackstodart This doesn't preserve the old heaps from before a retry yet, but it should fix the rest.

PR #973, opened 2026-05-07
Debugger records all old heaps and their parent heaps

jackstodart:
Changing the debugger from recording only heaps referenced in assumptions, to recording all heaps when they are created. Also records the statement or expression that caused the heap to be created, and any possible branch conditions.

Debugger now prints the parents and causes when printOldHeaps is on; it's a little verbose maybe, but it's off by default anyway.


src/main/scala/rules/Evaluator.scala
```diff
@@ -204,15 +206,12 @@ object evaluator extends EvaluationRules {
 
           val ve = pve dueTo InsufficientPermission(fa)
           v.heapSupporter.evalFieldAccess(s1, fa, tRcvr, eRcvr, ve, v1)((s2, snap, v2) => {
-            val (debugHeapName, debugLabel) = v1.getDebugOldLabel(s2, fa.pos, Some(magicWandSupporter.getEvalHeap(s2)))
+            val (_, debugLabel) = v1.getDebugOldLabel(s2, fa.pos, Some(magicWandSupporter.getEvalHeap(s2)))
```
marcoeilers: So this now assumes that the heap in which we're evaluating the field access has already been recorded, right? But I don't think that's necessarily the case, because of the scenario we I think briefly talked about but then got interrupted (or if there was some conclusion I don't remember it):
```
// heap h1
inhale acc(x.f) && x.f > 8 && acc(x.g)
// heap h2
```
``x.f`` is evaluated in a heap that is neither h1 nor h2, but an intermediate one that is h1 + the ``x.f`` permission. So I think here you'd get a new debug label that doesn't correspond to an existing heap, but the new heap is also not recorded.

src/main/scala/verifier/Verifier.scala
```diff
@@ -64,17 +64,62 @@ trait Verifier {
...
+  }
+
+  def recordOldHeap(s: State, parent: Heap, cause: ast.Exp, oldPCS: PathConditionStack): State = {
+    recordOldHeap(s, magicWandSupporter.getEvalHeap(s), parent, Right(cause), Some(oldPCS))
+  }
+
+  def recordOldHeap(s: State, heap: Heap, parent: Heap,
+                    cause: Either[ast.Stmt, ast.Exp],
+                    oldPCS: Option[PathConditionStack]): State = {
+    val childLabel = getDebugHeapLabel(s, Some(heap))
```
marcoeilers: Somehow this feels like we're mixing debug and old heaps in a way we maybe shouldn't. This will generate a debug heap labeleven when debugging is not enabled, e.g. when this is called for an ast.Label statement, right? 
Maybe we could use this as an opportunity to separate actual old heaps and debug heaps, and record the parents only for the debug heaps?

src/main/scala/state/State.scala
```diff
@@ -453,6 +469,11 @@ object State {
               Some(mergeHeap(heap1, cond1._1, cond1._2, heap2, cond2._1, cond2._2))
             }))
 
+            val oldHeapParents3 = Map.from(mergeMaps(oldHeapParents1, (), oldHeapParents2, ())
```
marcoeilers: My understanding is that since the heap parents already contain branch condition information, you can probably just take ``oldHeapParents1 ++ oldHeapParents2``?

src/main/scala/verifier/Verifier.scala
jackstodart: Yeah probably a good idea, I will re-rework this to separate the heaps then.

src/main/scala/state/State.scala
jackstodart: Roger that.

src/main/scala/rules/Evaluator.scala
jackstodart: OK I think this is finally fixed and now we record intermediate heaps when they are used (with either funcapp or field access).

src/main/scala/state/State.scala
```diff
@@ -136,6 +138,8 @@ final case class State(g: Store = Store(),
 
   val isLastRetry: Boolean = retryLevel == 0
 
+  val recordIntermediateHeaps: Boolean = temporaryHeapRecord.isDefined
```
AndreaKe: I'd rename this to something that indicates it's a boolean, e.g. isRecordingHeaps.

src/main/scala/state/State.scala
```diff
@@ -254,6 +286,18 @@ object State {
...
 
             val oldHeaps3 = oldHeaps1 ++ oldHeaps2
+            val debugOldHeaps3 = mergeMaps(debugOldHeaps1, (), debugOldHeaps2, ())(
+              (record, _) => Some(record))(
+              (record1, _, record2, _) =>
+                Some(record1.copy(intermediateHeaps = record1.intermediateHeaps ++ record2.intermediateHeaps))
+            )
+            val temporaryHeapRecord3 = (temporaryHeapRecord1, temporaryHeapRecord2) match {
+              case (Some((label1, cause1, pcs1, heaps1)), Some((label2, cause2, _, heaps2))) =>
+                if (label1 == label2 && cause1 == cause2) Some(label1, cause1, pcs1, heaps1 ++ heaps2) else None
```
AndreaKe: Is having different labels/causes something that can actually happen? Or does it indicate that something is wrong?

src/main/scala/supporters/functions/FunctionVerificationUnit.scala
```diff
@@ -260,23 +262,27 @@ trait DefaultFunctionVerificationUnitProvider extends VerifierComponent { v: Ver
...
+              val debugExp = if (debugOn) {
                 val e = ast.EqCmp(ast.Result(function.typ)(), body)(function.pos, function.info, function.errT)
                 val eNew = ast.EqCmp(ast.Result(function.typ)(), bodyNew.get)(function.pos, function.info, function.errT)
                 Some(DebugExp.createInstance(e, eNew))
               } else { None }
               decider.assume(BuiltinEquals(data.formalResult, tBody), debugExp)
-              consumes(s2, posts, false, postconditionViolated, v)((s3, _, _) => {
+              val s2a = if (debugOn) {
+                val tmp = v.finishKeyHeap(s2)
+                v.startKeyHeap(tmp, v.getDebugHeapLabel(tmp).getOrElse("missingFunctionHeap"), ExhalePost)
```
AndreaKe, line 283: Shouldn't there be a matching finishKeyHeap?

src/main/scala/supporters/MethodSupporter.scala
```diff
@@ -94,22 +98,29 @@ trait DefaultMethodVerificationUnitProvider extends VerifierComponent { v: Verif
...
                     Success()})})
             && {
-               executionFlowController.locally(s2a, v2)((s3, v3) =>  {
-                  exec(s3, body, v3)((s4, v4) =>
-                    consumes(s4, posts, false, postViolated, v4)((_, _, _) =>
-                      Success()))}) }  )})})
+               executionFlowController.locally(s2b, v2)((s3, v3) =>  {
+                  exec(s3, body, v3)((s4, v4) => {
+                    val currentLabel = v4.getDebugHeapLabel(s4).getOrElse("finalHeapMissing")
+                    val s4a = if (producer.debugOn) v4.startKeyHeap(s4, currentLabel, ExhalePost) else s4
```
AndreaKe, line 121: finishKeyHeap?

src/main/scala/verifier/Verifier.scala
```diff
@@ -58,24 +75,84 @@ trait Verifier {
     * @param pos the position of the current expression
     * @param h the heap to consider, if not the heap from state s
     * @return a pair containing the label of the given heap, and the label of the current expression in the given heap
-    */
-  def getDebugOldLabel(s: State, pos: ast.Position, h: Option[Heap] = None): (String, String) = {
+  def getDebugOldLabel(s: State, pos: ast.Position, h: Option[Heap] = None): String = {
```
AndreaKe: is this commented out on purpose? Can we remove it?

AndreaKe (commented):
I left some comments. Otherwise, LGTM.

src/main/scala/supporters/functions/FunctionVerificationUnit.scala
jackstodart: For function posts, that would only be after the function has been successfully verified so we would never display the resulting heap. But I could put it in there for completeness?

src/main/scala/supporters/MethodSupporter.scala
jackstodart: Same as for functions.

src/main/scala/state/State.scala
jackstodart: Agreed and changed.

src/main/scala/verifier/Verifier.scala
jackstodart: Oops, that very much should not still be there and so now it is no longer there.

src/main/scala/state/State.scala
jackstodart: This should not actually happen but I didn't want to cause crashes here if the debugger records heaps incorrectly. Added logger warning in this case. It would be very finicky to record the children heaps because I would have to put this condition inside the new `debugOldHeaps`.

src/main/scala/supporters/functions/FunctionVerificationUnit.scala
jackstodart: Decided not to add it because it would be a not used warning.

AndreaKe (approved):
LGTM

PR #976, opened 2026-05-26
Dependency Analysis

AndreaKe:
Adds the dependency analysis for Silicon. It is disabled by default. When enabled, it analyzes all (direct and indirect) semantic dependencies in the program. The dependencies of a proof obligation indicate which specifications, statements, and assumptions were used to prove that obligation. The analysis builds a dependency graph, which can be exported or queried using the included CLI tool.


src/main/scala/decider/Decider.scala
```diff
@@ -101,6 +112,13 @@ trait Decider {
   def setPcs(other: PathConditionStack): Unit
 
   def statistics(): Map[String, String]
+
+  var dependencyAnalyzer: DependencyAnalyzer
+  def initDependencyAnalyzer(member: Member, preambleNodes: Iterable[DependencyAnalysisNode]): Unit
```
marcoeilers: I'm a bit surprised that this is in the Decider.

src/main/scala/decider/Decider.scala
```diff
@@ -49,31 +52,39 @@ trait Decider {
...
 
-  def assume(t: Term, e: ast.Exp, finalExp: ast.Exp): Unit
-  def assume(t: Term, e: Option[ast.Exp], finalExp: Option[ast.Exp]): Unit
-  def assume(t: Term, debugExp: Option[DebugExp]): Unit
-  def assume(terms: Seq[Term], debugExps: Option[Seq[DebugExp]]): Unit
-  def assumeDefinition(t: Term, debugExp: Option[DebugExp]): Unit
-  def assume(assumptions: Iterable[(Term, Option[DebugExp])]): Unit
-  def assume(assumptions: InsertionOrderedSet[(Term, Option[DebugExp])], enforceAssumption: Boolean = false, isDefinition: Boolean = false): Unit
-  def assume(terms: Iterable[Term], debugExp: Option[DebugExp], enforceAssumption: Boolean): Unit
+  def registerChunk[CH <: GeneralChunk](buildChunk: Term => CH, perm: Term, analysisInfo: AnalysisInfo, isExhale: Boolean): CH
```
marcoeilers: Should this be in the Decider? It doesn't feel very Decider-like.

src/main/scala/decider/Decider.scala
```diff
@@ -49,31 +52,39 @@ trait Decider {
...
+  def registerDerivedChunk[CH <: GeneralChunk](sourceChunks: Set[Chunk], buildChunk: Term => CH, perm: Term, analysisInfo: AnalysisInfo, isExhale: Boolean, createLabel: Boolean=true): CH
+  def wrapWithDependencyAnalysisLabel(term: Term, sourceChunks: Iterable[Chunk] = Set.empty, sourceTerms: Iterable[Term] = Set.empty): Term
+  def isPathInfeasible: Boolean
+
+  def assume(t: Term, e: Option[ast.Exp], finalExp: Option[ast.Exp], analysisInfos: DependencyAnalysisInfos): Unit
+  def assume(t: Term, debugExp: Option[DebugExp], analysisInfos: DependencyAnalysisInfos): Unit
```
marcoeilers, line 69: I'm wondering if we at some point should summarize the additional information for specific modes (debugging, dependency analysis, Joao has a branch where he is tracking query types https://github.com/viperproject/silicon/pull/975) in some sort of info argument

src/main/scala/decider/Decider.scala
```diff
@@ -270,36 +310,67 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
       }
     }
 
+    def registerChunk[CH <: GeneralChunk](buildChunk: Term => CH, perm: Term, analysisInfo: AnalysisInfo, isExhale: Boolean): CH = {
+      registerDerivedChunk[CH](Set.empty, buildChunk, perm, analysisInfo, isExhale)
+    }
+
+    def registerDerivedChunk[CH <: GeneralChunk](sourceChunks: Set[Chunk], buildChunk: Term => CH, perm: Term, analysisInfo: AnalysisInfo, isExhale: Boolean, createLabel: Boolean=true): CH = {
+      if(!isDependencyAnalysisEnabled)
```
marcoeilers: Wait so this does something even if the analysis is disabled?

src/main/scala/decider/Decider.scala
```diff
@@ -270,36 +310,67 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
...
+        return buildChunk(perm)
+
+      val labelNodeOpt = getOrCreateAnalysisLabelNode()
+
+      if(isExhale)
```
marcoeilers: Silicon uses ``if (`` (with a space) throughout

src/main/scala/decider/Decider.scala
```diff
@@ -341,7 +446,10 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
       })
     }
 
-    private def assumeWithoutSmokeChecks(terms: InsertionOrderedSet[Term], isDefinition: Boolean = false) = {
+    private def assumeWithoutSmokeChecks(termsWithLabel: InsertionOrderedSet[(Term, String)], isDefinition: Boolean = false): None.type = {
+      if (isPathInfeasible) return None
```
marcoeilers: I don't understand why we are returning None

src/main/scala/decider/Decider.scala
```diff
@@ -353,26 +461,79 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
...
+
       val timeout = if (isAssert) Verifier.config.assertTimeout.toOption else Verifier.config.checkTimeout.toOption
-      prover.check(timeout) == Unsat
+      val result = prover.check(timeout, label) == Unsat
+
+      if(result){
+        checkNode foreach dependencyAnalyzer.addAssertionNode
+        dependencyAnalyzer.processUnsatCoreAndAddDependencies(prover.getLastUnsatCore, label)
+        val infeasibleNodeId = dependencyAnalyzer.addInfeasibilityNode(!isAssert, analysisInfos)
+//        Adding the following line would be UNSOUND! Unsoundness is introduced when infeasibility is introduced while executing a package statements and pontentially in other cases as well.
```
marcoeilers: Should this comment stay around?

src/main/scala/decider/PathConditions.scala
```diff
@@ -100,6 +106,7 @@ private class PathConditionStackLayer
   private var _globalDefiningAssumptionDebugExps: InsertionOrderedSet[DebugExp] = InsertionOrderedSet.empty
   private var _nonGlobalDefiningAssumptionDebugExps: InsertionOrderedSet[DebugExp] = InsertionOrderedSet.empty
   private var _declarations: InsertionOrderedSet[Decl] = InsertionOrderedSet.empty
+  private var _analysisLabels: InsertionOrderedSet[Term] = InsertionOrderedSet.empty
```
marcoeilers, line 113: I'm generally wondering if it would be possible to somehow separate the dependency analysis code more from the rest. For example, you could extend the interface here with the methods you need, and then implement them only in a subclass (and have noop-implementations in the standard implementation)?

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -294,7 +296,8 @@ class Z3ProverAPI(uniqueId: String,
     cleanTerm
   }
 
-  def assert(goal: Term, timeout: Option[Int]): Boolean = {
+  // TODO ake: label
```
marcoeilers: !!

src/main/scala/decider/Z3ProverAPI.scala
```diff
@@ -402,6 +405,8 @@ class Z3ProverAPI(uniqueId: String,
     }
   }
 
+  def getLastUnsatCore: String = "" // TODO ake
```
marcoeilers: !!

src/main/scala/dependencyAnalysis/cliTool/BenchmarkDependencyAnalysisCliExtension.scala
```diff
@@ -0,0 +1,249 @@
...
+import java.time.format.DateTimeFormatter
+import scala.io.Source
+import scala.io.StdIn.readLine
+import scala.util.matching.Regex
+
+class BenchmarkDependencyAnalysisCliExtension(override val interpreter: DependencyGraphInterpreter[Final], program: ast.Program) extends DependencyAnalysisCliToolExtension {
+
+	override val name: String = "Benchmark Features"
+	override val commands: List[DependencyAnalysisCliCommand] = List(
+																																new PerformanceBenchmarkCommand,
```
marcoeilers: Indentation looks a bit weird

src/main/scala/dependencyAnalysis/cliTool/BenchmarkDependencyAnalysisCliExtension.scala
```diff
@@ -0,0 +1,249 @@
...
+
+				val isSound = groundTruthLabels.diff(actualLabelInReportedDeps).isEmpty
+				val imprecise = actualLabelInReportedDeps.diff(groundTruthLabels)
+
+				assert(!isSound || groundTruthLabels.size + imprecise.size == actualLabelInReportedDeps.size, s"Imprecision calculation is wrong.")
+				assert(actualLabelInReportedDeps.size <= callGraphLabels.size, "Call graph size is smaller than reported dependencies.")
+
+				addOutput(bw, s"$assertionLabel,${if (isSound) "YES" else "NO"},${groundTruthLabels.size},${actualLabelInReportedDeps.size},${imprecise.size},${callGraphLabels.size},${durationMs}ms,${noise.size}")
+
+				//      println(s"Queried:\n\t${getSourceInfoString(queriedAssertions)}")
```
marcoeilers: Drop?

src/main/scala/interfaces/state/Chunks.scala
```diff
@@ -7,47 +7,121 @@
...
+  protected def permScale(perm: Term, permExp: Option[ast.Exp]): GeneralChunk
+
+  protected def substitute(terms: silicon.Map[Term, Term]): GeneralChunk
 
   val permExp: Option[ast.Exp]
 }
 
+object GeneralChunk {
+  def applyCondition(chunk: GeneralChunk, newCond: Term, newCondExp: Option[ast.Exp], analysisInfo: AnalysisInfo): GeneralChunk = {
+    analysisInfo.decider.registerDerivedChunk[GeneralChunk](Set(chunk), {_ =>
```
marcoeilers: So this really seems more like a chunk factory then, that also registers things. I think that should be named differently.
Or... does it have to be set up this way? Could you not create the chunk separately and then register it afterwards?

src/main/scala/resources/NonQuantifiedPropertyInterpreter.scala
```diff
@@ -120,8 +122,9 @@ class NonQuantifiedPropertyInterpreter(heap: Iterable[Chunk], verifier: Verifier
                                     otherwise: PropertyExpression[K],
                                     info: Info): (Term, Option[ast.Exp]) = {
     val conditionTerm = buildPathCondition(condition, info)._1
-    if (verifier.decider.check(conditionTerm, Verifier.config.checkTimeout())) {
-      buildPathCondition(thenDo, info)
+    if (verifier.decider.check(conditionTerm, Verifier.config.checkTimeout(), DependencyAnalysisInfos.create(s"property interpreter: ${conditionTerm.toString}", DependencyType.Internal))) {
+      val (t, e) = buildPathCondition(thenDo, info)
+      (verifier.decider.wrapWithDependencyAnalysisLabel(t, Set.empty, Set(conditionTerm)), e) // TODO ake: causes imprecision!
```
marcoeilers: Could define a shorthand for that

src/main/scala/rules/Brancher.scala
```diff
@@ -97,7 +109,7 @@ object brancher extends BranchingRules {
     var noOfErrors = 0
 
     val elseBranchVerificationTask: Verifier => VerificationResult =
-      if (executeElseBranch) {
+      if (executeElseBranch || Verifier.config.disableInfeasibilityChecks()) {
```
marcoeilers: There is a variable ``skipPathFeasibilityCheck``, maybe you could reuse that?

src/main/scala/rules/ChunkSupporter.scala
```diff
@@ -214,10 +235,10 @@ object chunkSupporter extends ChunkSupportRules {
   def produce(s: State, h: Heap, ch: NonQuantifiedChunk, v: Verifier)
              (Q: (State, Heap, Verifier) => VerificationResult)
              : VerificationResult = {
-
+    val analysisInfos = DependencyAnalysisInfos.DefaultDependencyAnalysisInfos.withSource(StringAnalysisSourceInfo("produce", ast.NoPosition)).withDependencyType(DependencyType.Internal)
```
marcoeilers: That feels like it could be abbreviated :D

src/main/scala/rules/Evaluator.scala
```diff
@@ -36,11 +38,11 @@ import viper.silver.utility.Common.Rational
  */
 
 trait EvaluationRules extends SymbolicExecutionRules {
-  def evals(s: State, es: Seq[ast.Exp], pvef: ast.Exp => PartialVerificationError, v: Verifier)
+  def evals(s: State, es: Seq[ast.Exp], pvef: ast.Exp => PartialVerificationError, v: Verifier, analysisInfos: DependencyAnalysisInfos)
            (Q: (State, List[Term], Option[List[ast.Exp]], Verifier) => VerificationResult)
            : VerificationResult
 
-  def eval(s: State, e: ast.Exp, pve: PartialVerificationError, v: Verifier)
+  def eval(s: State, e: ast.Exp, pve: PartialVerificationError, v: Verifier, analysisInfos: DependencyAnalysisInfos)
```
marcoeilers, line 45: Make sure this is consistent: Verifier used to be last everywhere, now AnalysisInfos are always after the Verifier (I'm not saying it's not consistent, I just didn't pay attention until now)

src/main/scala/rules/Evaluator.scala
```diff
@@ -61,40 +64,41 @@ object evaluator extends EvaluationRules {
...
-    eval3(s, e, pve, v)((s1, t, eNew, v1) => {
+    val analysisInfos1 = v.decider.handleAndGetUpdatedAnalysisInfos(analysisInfos, e.info, e)
+
+    eval3(s, e, pve, v, analysisInfos1)((s1, t, eNew, v1) => {
       v1.symbExLog.closeScope(sepIdentifier)
       Q(s1, t, eNew, v1)})
   }
 
-  def eval3(s: State, e: ast.Exp, pve: PartialVerificationError, v: Verifier)
+  def eval3(s: State, e: ast.Exp, pve: PartialVerificationError, v: Verifier, analysisInfos: DependencyAnalysisInfos)
```
marcoeilers, line 98: So the analysisInfos couldn't possible be part of the State, could they?

src/main/scala/state/Chunks.scala
```diff
@@ -45,45 +59,76 @@ case class BasicChunk(resourceID: BaseID,
...
+    GeneralChunk.permPlus(chunk, perm, permExp, analysisInfo).asInstanceOf[QuantifiedBasicChunk]
+}
+
 sealed trait QuantifiedBasicChunk extends QuantifiedChunk {
   override val id: ChunkIdentifer
-  override def applyCondition(newCond: Term, newCondExp: Option[ast.Exp]): QuantifiedBasicChunk
-  override def permMinus(perm: Term, permExp: Option[ast.Exp]): QuantifiedBasicChunk
-  override def permPlus(perm: Term, permExp: Option[ast.Exp]): QuantifiedBasicChunk
-  override def withSnapshotMap(snap: Term): QuantifiedBasicChunk
+  override protected def applyCondition(newCond: Term, newCondExp: Option[ast.Exp]): QuantifiedBasicChunk
```
marcoeilers: Aw, this was such a nice interface though :(

marcoeilers (commented):
So I had a rough look and left some nitpicky comments and some very general questions; let's discuss those at some point.

src/main/scala/rules/Evaluator.scala
AndreaKe: If I put it into the state, it would be passed to the continuation pretty much everywhere, right? That shouldn't happen for the analysisInfos

src/main/scala/interfaces/state/Chunks.scala
AndreaKe: I guess you're right that it's a factory. This is probably the ugliest part of the implementation but I'm not sure how to improve it. During registering, the permissions are "wrapped" by an analysis label (e.g. label2? FullPerm : NoPerm). Since I can't modify existing chunks, I need to do it at creation.

src/main/scala/decider/Decider.scala
AndreaKe: Me neither. The function seems to always return None... I removed it

src/main/scala/rules/Evaluator.scala
AndreaKe: I double-checked that it is consistent.

src/main/scala/decider/Z3ProverAPI.scala
AndreaKe: I added a validation for the config flags to ensure that the Z3ProverAPI is not used in combination with the dependency analysis. Hence, there is nothing to do here. I removed the comments.

src/main/scala/state/Chunks.scala
AndreaKe: These functions can still be called. Just not on the Chunks directly but via the ChunkFactory. The DependencyAnalysisAwareChunkFactory ensures that the chunks are properly represented in the dependency graph, which is why one must always create chunks via the factory.

src/main/scala/decider/Decider.scala
```diff
@@ -270,36 +288,33 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
       }
     }
 
+    def isPathInfeasible: Boolean = Verifier.config.disableInfeasibilityChecks() && pcs.isPathInfeasible
```
marcoeilers: I'm not sure about the name. This seems to not actually check if the current path is infeasible, which is what a caller would probably expect.

src/main/scala/resources/NonQuantifiedPropertyInterpreter.scala
```diff
@@ -120,8 +123,9 @@ class NonQuantifiedPropertyInterpreter(heap: Iterable[Chunk], verifier: Verifier
                                     otherwise: PropertyExpression[K],
                                     info: Info): (Term, Option[ast.Exp]) = {
     val conditionTerm = buildPathCondition(condition, info)._1
-    if (verifier.decider.check(conditionTerm, Verifier.config.checkTimeout())) {
-      buildPathCondition(thenDo, info)
+    if (verifier.decider.check(conditionTerm, Verifier.config.checkTimeout(), DependencyAnalysisInfos.create(s"property interpreter: ${conditionTerm.toString}", DependencyType.Internal))) {
+      val (t, e) = buildPathCondition(thenDo, info)
+      (DependencyAnalyzer.wrapWithDependencyAnalysisLabel(verifier.decider, t, Set.empty, Set(conditionTerm)), e) // TODO ake: causes imprecision!
```
marcoeilers, line 127: Is this TODO a problem?

src/main/scala/Config.scala
```diff
@@ -691,6 +692,48 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+    noshort = true
+  )
+
+  val enableDependencyAnalysisDebugging: ScallopOption[Boolean] = opt[Boolean]("enableDependencyAnalysisDebugging",
+    descr = "Enable debugging for dependency analysis mode",
+    default = Some(false),
+    noshort = true
+  )
+
+  val disableInfeasibilityChecks: ScallopOption[Boolean] = opt[Boolean]("disableInfeasibilityChecks",
```
marcoeilers: What's the purpose of this one? Doesn't hurt to have of course, but since it seems to be independent of other options, I'm surprised it exists.

src/main/scala/Config.scala
```diff
@@ -738,6 +781,45 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+    case (_, Some(false)) => Right(())
+    case (Some(true), Some(true)) =>
+      Left(s"Option ${enableDependencyAnalysis.name} is not supported in combination with ${parallelizeBranches.name}")
+    case other =>
+      sys.error(s"Unexpected combination: $other")
+  }
+
+  validateOpt(rawProverArgs, enableDependencyAnalysis) {
+    case (_, Some(false)) => Right(())
+    case (Some(args), Some(true)) if args.contains("proof=true") && args.contains("unsat-core=true") => Right(())
```
marcoeilers: pretty sure Z3 accepts PROOF=TRUE etc., so this should handle different capitalization

src/main/scala/Config.scala
```diff
@@ -691,6 +692,48 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
     noshort = true
   )
 
+  val startDebuggerAutomatically: ScallopOption[Boolean] = opt[Boolean]("startDebuggerAutomatically",
+    descr = "Starts the debugging mode automatically after verification completes",
```
marcoeilers: What happens otherwise if debugging is enabled but this is off?

src/main/scala/supporters/DependencyAnalysisAwareMethodVerificationUnitProvider.scala
```diff
@@ -0,0 +1,42 @@
...
+import viper.silver.dependencyAnalysis._
+
+trait DependencyAnalysisAwareMethodVerificationUnitProvider extends DefaultMethodVerificationUnitProvider { v: DependencyAnalysisAwareVerifier =>
+  override def methodSupporter: MethodSupporter = DependencyAnalysisAwareMethodSupporter
+
+  private object DependencyAnalysisAwareMethodSupporter extends MethodSupporter {
+
+    override def verify(sInit: State, method: Method): Seq[VerificationResult] = {
+
+      val presAssertionNodeForJoin = method.pres.flatMap(_.topLevelConjuncts).map(pc => SimpleAssertionNode(True, AnalysisSourceInfo.createAnalysisSourceInfo(pc), AssumptionType.Precondition, SimpleDependencyAnalysisMerge(AnalysisSourceInfo.createAnalysisSourceInfo(pc)), List(SimpleDependencyAnalysisJoin(AnalysisSourceInfo.createAnalysisSourceInfo(pc), JoinType.Sink, EdgeType.Up)), method.name))
```
marcoeilers: This line looks very, very, very long.

src/main/scala/supporters/functions/FunctionData.scala
```diff
@@ -296,25 +344,33 @@ class FunctionData(val programFunction: ast.Function,
...
            Seq(Trigger(functionApplication)) ++ actualPredicateTriggers)
 
-      Forall(arguments, body, allTriggers)})
+      (Forall(arguments, body, allTriggers),
+        bodyAnalysisInfo)
+    })
   }
 
-  lazy val bodyPreconditionPropagationAxiom: Seq[Term] = {
+	lazy val definitionalAxiom: Option[Term] = definitionalAxiomWithInfo.map(_._1)
```
marcoeilers: I haven't checked throughout but this looks like the indentation is wrong

src/main/scala/state/Utils.scala
```diff
@@ -6,6 +6,8 @@
 
 package viper.silicon.state
 
+import viper.silicon.state.chunks.{BasicChunk, QuantifiedFieldChunk}
```
marcoeilers, line 9: Probably not needed, since nothing else changed?

src/main/scala/state/SnapshotMapCache.scala
```diff
@@ -6,7 +6,9 @@
 
 package viper.silicon.state
 
+import viper.silicon.state.chunks.QuantifiedBasicChunk
```
marcoeilers, line 9: Probably not needed?

marcoeilers (approved):
Some comments here and there, otherwise LGTM

src/main/scala/resources/NonQuantifiedPropertyInterpreter.scala
AndreaKe: It's not really a problem. The analysis results are a bit imprecise because of the way I wrap the assumptions here but it is sound. I plan to optimize this in the future.

src/main/scala/decider/Decider.scala
AndreaKe: Renamed to isPathMarkedInfeasible. Also in the path condition stack and anywhere else this name was used.

src/main/scala/Config.scala
AndreaKe: I renamed it to analyzeInfeasiblePaths. I think that explains is better. It basically forces verification to step through infeasible paths (without actually emitting assumptions or assertions) such that we can collect analysis information required to make our approach sound.

src/main/scala/Config.scala
AndreaKe: I removed this. i think it is a leftover from the early days where the dependency analysis used part of the debug information. Now the dependency analysis is independent of the debugging information. Hence, this config flag is obsolete.

src/main/scala/supporters/functions/FunctionData.scala
AndreaKe: IntelliJ using tabs instead of spaces again... Fixed now.

src/main/scala/state/Utils.scala
AndreaKe: It's needed because I moved the Chunk classes and interfaces to a new package.

src/main/scala/decider/Decider.scala
AndreaKe: I'll discuss this with Joao once he's back.

src/main/scala/debugger/SiliconDebugger.scala
```diff
@@ -270,7 +270,7 @@ class SiliconDebugger(verificationResults: List[VerificationResult],
   }
 
   private def initVerifier(obl: ProofObligation, proverName: String, userArgsString: Option[String]): ProofObligation = {
-    val v = new WorkerVerifier(this.mainVerifier, obl.v.uniqueId, NoopReporter, false)
+    val v = new WorkerVerifier(this.mainVerifier, "debugger_01", NoopReporter, false)
```
jcp19: is there an issue if we always create a verifier with the same id? (especially if more than one WorkerVerifier may be created throughout the execution)

src/main/scala/decider/Decider.scala
jcp19, line 39: So, we are making extensive changes to the decider so that we always propagate the analysisInfos data. Is it ever the case that this object is propagated through every call without changing it? If so, do we really need to make such extensive changes to the decider? Instead, couldn't we just either
(1) extend the decider with a single function `analysisInfos` and call it only where this info is relevant, or
(2) implement a subtrait of the decider for "dependency aware decider" that adds the necessary instrumentation to every decider by performing super calls?

jcp19, line 335: why does this become an `InsertionOrderedSet`?

src/main/scala/decider/Decider.scala
```diff
@@ -49,31 +55,35 @@ trait Decider {
...
-  def assume(terms: Iterable[Term], debugExp: Option[DebugExp], enforceAssumption: Boolean): Unit
+  def debuggerAssume(terms: Iterable[Term], de: DebugExp)
+
+  def isPathMarkedInfeasible: Boolean
+  def handleInfeasiblePath(hasAssertions: Boolean, hasAssumptions: Boolean, analysisInfos: DependencyAnalysisInfos): Unit
 
-  def check(t: Term, timeout: Int): Boolean
+  def assume(t: Term, e: Option[ast.Exp], finalExp: Option[ast.Exp], analysisInfos: DependencyAnalysisInfos): Unit
+  def assume(t: Term, debugExp: Option[DebugExp], analysisInfos: DependencyAnalysisInfos): Unit
+  def assume(assumptions: Iterable[Term], debugExps: Option[Iterable[DebugExp]], description: String, enforceAssumption: Boolean, analysisInfos: DependencyAnalysisInfos): Unit
```
jcp19, line 70: what is `enforceAssumption` supposed to mean?

src/main/scala/decider/Decider.scala
```diff
@@ -309,26 +332,45 @@ trait DefaultDeciderProvider extends VerifierComponent { this: Verifier =>
...
 
-    def assume(terms: Iterable[Term], debugExp: Option[DebugExp], enforceAssumption: Boolean): Unit = {
+    def assume(assumptions: Iterable[Term], debugExps: Option[Iterable[DebugExp]], description: String, enforceAssumption: Boolean, analysisInfos: DependencyAnalysisInfos): Unit = {
+      val debugExp = Option.when(debugExps.isDefined)(DebugExp.createInstance(description, InsertionOrderedSet(debugExps.get)))
+
+      val filteredTerms =
+        if (enforceAssumption) assumptions
+        else assumptions filterNot isKnownToBeTrue
+
+      if (filteredTerms.isEmpty) return
```
jcp19: return expressions have a somewhat funky semantics in scala due to non-local semantics. In general, I think it would be better to restrcutre the code such that the early return is not used

src/main/scala/decider/DependencyAnalysisAwareDeciderProvider.scala
```diff
@@ -0,0 +1,185 @@
...
+import viper.silicon.dependencyAnalysis._
+import viper.silicon.state.chunks.{Chunk, GeneralChunk}
+import viper.silicon.state.terms.{False, Term, True}
+import viper.silicon.verifier.{DependencyAnalysisAwareVerifier, Verifier}
+import viper.silver.ast
+import viper.silver.ast.Member
+
+trait DependencyAnalysisDeciderFeatures {
+
+  def isDependencyAnalysisEnabled: Boolean
```
jcp19, line 20: is this flag meant to be set once for the entire execution of the program, or can it change at runtime?

src/main/scala/decider/DependencyAnalysisAwareDeciderProvider.scala
```diff
@@ -0,0 +1,185 @@
...
+
+import viper.silicon.common.collections.immutable.InsertionOrderedSet
+import viper.silicon.dependencyAnalysis._
+import viper.silicon.state.chunks.{Chunk, GeneralChunk}
+import viper.silicon.state.terms.{False, Term, True}
+import viper.silicon.verifier.{DependencyAnalysisAwareVerifier, Verifier}
+import viper.silver.ast
+import viper.silver.ast.Member
+
+trait DependencyAnalysisDeciderFeatures {
```
jcp19: this name is a bit awkward. Maybe `DependencyAnalysisHandler` would be more suitable? Also, does this have anything to do with the Decider at all?

src/main/scala/dependencyAnalysis/DependencyAnalysisInfos.scala
```diff
@@ -0,0 +1,146 @@
...
+import viper.silver.ast
+import viper.silver.ast._
+import viper.silver.dependencyAnalysis.AssumptionType.AssumptionType
+import viper.silver.dependencyAnalysis._
+
+/**
+ * Stores all information about the currently evaluated statement/expression such that the dependency analysis can
+ * correctly add nodes and edges to the graph.
+ */
+case class DependencyAnalysisInfos(sourceInfos: List[DependencyAnalysisSourceInfo], dependencyTypes: List[DependencyTypeInfo], mergeInfos: List[DependencyAnalysisMergeInfo], joinInfos: List[DependencyAnalysisJoinInfo], nodes: List[ast.Node], analysisEnabled: Boolean = true) {
```
jcp19, line 67: line too long, would be good to document what each node represents

src/main/scala/dependencyAnalysis/DependencyAnalysisInfos.scala
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+
+  private def getDebugInfo: String = {
+    val sourceInfo = sourceInfos.headOption.map("source info: " + _.toString + " ").getOrElse("")
+    val nodeInfo = if (nodes.nonEmpty) "nodes: " + nodes.map(getNodeInfo).mkString(", ") else ""
+    s"$sourceInfo$nodeInfo"
+  }
+
+  def getSourceInfo: DependencyAnalysisSourceInfo = {
```
jcp19, line 129: many of these functions return something from the first elements of the respective lists. why is that? what invariant is this relying on?

src/main/scala/dependencyAnalysis/DependencyAnalysisInfos.scala
jcp19, line 67: Nit: given that the packages already provide some kind of namespacing for names, you don't need to prefix every name with "DependencyAnalysis"

src/main/scala/dependencyAnalysis/DependencyAnalysisInfos.scala
```diff
@@ -0,0 +1,146 @@
...
+    this.withSource(sourceInfo).withDependencyType(assumptionType)
+}
+
+object DependencyAnalysisInfos {
+  val DefaultInfos = DependencyAnalysisInfos(List.empty, List.empty, List.empty, List.empty, List.empty)
+}
+
+case class DependencyAnalysisAxiomInfo(analysisInfos: DependencyAnalysisInfos, memberStr: String) {
+
+}
```
jcp19: drop empty parentheses

src/main/scala/dependencyAnalysis/DependencyAnalysisNode.scala
```diff
@@ -0,0 +1,157 @@
...
+import viper.silver.dependencyAnalysis.AssumptionType.AssumptionType
+import viper.silver.dependencyAnalysis._
+
+
+trait DependencyAnalysisNode {
+
+  /**
+   * The unique node id, which is also given to the SMT solver such that unsat cores can be mapped back to dependency nodes.
+   */
+  val _id: Option[Int]
```
jcp19: shouldn't this be private?

src/main/scala/dependencyAnalysis/DependencyAnalysisNode.scala
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * The program member (method, function) that the node belongs to
+   */
+  val memberStr: String
+
+  /**
+   * The assumes or asserted Silicon term. Currently, only used for debugging purposes.
+   */
+  val term: Term
+  def getTerm: Term = term
```
jcp19: why do we have both?

src/main/scala/dependencyAnalysis/DependencyAnalysisNode.scala
```diff
@@ -0,0 +1,157 @@
...
+  def getAssertFailedNode: GeneralAssertionNode
+}
+
+// this is not strictly needed anymore but storing the chunk and label node is useful for debugging purposes
+trait ChunkAnalysisInfo {
+  val chunk: Chunk
+  val labelNode: LabelNode
+}
+
+class SimpleAssumptionNode(override val term: Term, description: Option[String], override val sourceInfo: DependencyAnalysisSourceInfo, override val assumptionType: AssumptionType, override val mergeInfo: DependencyAnalysisMergeInfo, override val joinInfos: List[SimpleDependencyAnalysisJoin], override val memberStr: String, override val _id: Option[Int]=None) extends GeneralAssumptionNode {
```
jcp19: line too long

src/main/scala/dependencyAnalysis/DependencyAnalysisResult.scala
```diff
@@ -0,0 +1,16 @@
...
+// Copyright (c) 2011-2026 ETH Zurich.
+
+package viper.silicon.dependencyAnalysis
+
+import viper.silicon.dependencyAnalysis.graphInterpretation.DependencyGraphInterpreter
+import viper.silver.ast.Program
+
+case class DependencyAnalysisResult(program: Program, fullDependencyGraphInterpreter: DependencyGraphInterpreter[Final]) {
+
+  def getFullDependencyGraphInterpreter: DependencyGraphInterpreter[Final] = fullDependencyGraphInterpreter
```
jcp19: isn't this redundant? we might as well just use the field name

src/main/scala/dependencyAnalysis/DependencyAnalysisNode.scala
```diff
@@ -0,0 +1,157 @@
...
+  override def getNodeString: String = "assume " + term.toString + description.map(" (" + _ + ")").getOrElse("")
+}
+
+class AxiomAssumptionNode(override val term: Term, description: Option[String], override val sourceInfo: DependencyAnalysisSourceInfo, override val assumptionType: AssumptionType, override val mergeInfo: DependencyAnalysisMergeInfo, override val joinInfos: List[SimpleDependencyAnalysisJoin], override val memberStr: String, override val _id: Option[Int]=None) extends GeneralAssumptionNode {
```
jcp19: any reason why these are not case classes?

src/main/scala/dependencyAnalysis/DependencyGraph.scala
```diff
@@ -0,0 +1,398 @@
...
+
+  /**
+   * Helper function used to ensure uniqueness of all dependency node ids.
+   */
+  def nextId(): Int = {
+    idCounter.getAndIncrement()
+  }
+}
+
+trait DependencyGraphState
```
jcp19: shouldn't it be sealed?

src/main/scala/Config.scala
```diff
@@ -691,6 +692,42 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
     noshort = true
   )
 
+  val enableUnsatCores: ScallopOption[Boolean] = opt[Boolean]("enableUnsatCores",
+    descr = "Enable UNSAT cores",
+    default = Some(false),
+    noshort = true
+  )
```
jcp19, line 699: is there a reason why we would have unsat cores enabled but not dependency analysis? Is it just for benchmarking? Might make sense to mention that here. "Compute UNSAT cores. Useful for evaluating the overhead of this feature in an isolated maner."

src/main/scala/Config.scala
```diff
@@ -691,6 +692,42 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+
+  val enableDependencyAnalysis: ScallopOption[Boolean] = opt[Boolean]("enableDependencyAnalysis",
+    descr = "Enable the verification dependency analysis",
+    default = Some(false),
+    noshort = true
+  )
+
+  val disableDependencyAnalysisJoinPrecisionOpt: ScallopOption[Boolean] = opt[Boolean]("disableDependencyAnalysisJoinPrecisionOpt",
+    descr = "Disable the precision optimizations regarding edges across method boundaries.",
```
jcp19:
```suggestion
    descr = "Use a less precise but faster algorithm for computing dependencies.",
```

src/main/scala/Config.scala
```diff
@@ -691,6 +692,42 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+    descr = "Enable analysis of infeasible paths by making the verification step through all paths (even if provably infeasible)",
+    default = Some(false),
+    noshort = true
+  )
+
+  val dependencyAnalysisMode: ScallopOption[String] = opt[String]("dependencyAnalysisMode",
+    descr = "Set verification dependency analysis commands (separated by ;) to be executed after verification. Available are `interactive` and all commands supported by the interactive CLI tool",
+    default = None,
+    noshort = true
+  )
```
jcp19: I think this line is a bit confusing.
what happens if I write "dep 10; interactive; interactive"?

also, "available are" sounds weird 😅 It breaks the SVO structure in english. "The available commands are 'interactive' and ..." sounds more natural to me

src/main/scala/Config.scala
```diff
@@ -691,6 +692,42 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+    descr = "Enable debugging for the verification dependency analysis mode",
+    default = Some(false),
+    noshort = true
+  )
+
+  val analyzeInfeasiblePaths: ScallopOption[Boolean] = opt[Boolean]("analyzeInfeasiblePaths",
+    descr = "Enable analysis of infeasible paths by making the verification step through all paths (even if provably infeasible)",
+    default = Some(false),
+    noshort = true
+  )
```
jcp19: hmm, I would probably set this as the default behaviour (it seems that we always use the two together, and disabling it leads to potentially unsound results for the dependency analysis). This means that we should probably rename this to "disableInfeasiblePaths". We should also explain why people might want to use it with a sentence of the kind "This flag makes the dependency analysis faster by skipping the analysis of dead paths, but it may produce unsound results where the dependencies of a proof obligation in an infeasible branch are not completely listed."

src/main/scala/Config.scala
```diff
@@ -691,6 +692,42 @@ class Config(args: Seq[String]) extends SilFrontendConfig(args, "Silicon") {
...
+    descr = "Enable UNSAT cores",
+    default = Some(false),
+    noshort = true
+  )
+
+  val enableDependencyAnalysis: ScallopOption[Boolean] = opt[Boolean]("enableDependencyAnalysis",
+    descr = "Enable the verification dependency analysis",
+    default = Some(false),
+    noshort = true
+  )
```
jcp19: are we always expected to provide a mode togehter with this flag? That makes this flag redundant. In that case, we could drop it and just rename the `dependencyAnalysisMode` to `depAnalysis`

jcp19: (I guess the point I am making in this comment and others is that the typical use of the dependency analysis should involve only one extra flag, not three like we currently have, which makes it hard to remember which combination of flags we need to pass)

src/main/scala/decider/Decider.scala
AndreaKe: I don't see how I could get rid of the analysisInfo argument. While it might happen that several consecutive calls to assume/assert have the same analysisInfo, this does not hold in general. Usually, the info is specific to a single call to assume and assert, so passing it as an argument makes most sense to me. Let me know when you have an idea on how to improve it.

As a side note, I used to store the analysisInfos as a field in the decider and thus didn't have this extra argument (neither in the decider, nor in the evaluator, executor, producer, ...). However, I had to implement several hacks to keep the info consistent and as a result assumptions were associated with a wrong source/dependency type. That's why I switched to the current design at some point.

src/main/scala/decider/Decider.scala
AndreaKe: If `enforceAssumption=true`, all assumptions are passed to the SMT solver, regardless of whether they are already known to be true. This is a feature that existed already and I didn't change anything about it.

src/main/scala/decider/DependencyAnalysisAwareDeciderProvider.scala
AndreaKe: It can change at runtime. It mostly depends on the member that is verified, e.g. we disable analysis of builtin Gobra functions & methods.

src/main/scala/decider/Decider.scala
AndreaKe: Not sure, this was already there...

PR #849, opened 2024-06-12
Fix multiple `apply` of magic wands with quantified expressions

manud99:
In my first PR #836 I didn't fix the `apply` of magic wands with quantified expressions. In this PR I address this issue and propose a fix for all combinations of magic wands with quantified expression that I could find in silver's test suite. This PR includes:

* Using MagicWandSnapFunctions (MWSF) also for quantified magic wands.
* Analyze and rewrite all path conditions that were created during packaging.

Test suite can be found in PR viperproject/silver#805

PR #985, opened 2026-06-29
Encoding quantified wands using MWSFs

marcoeilers:
This PR replaces the encoding of magic wand snapshots for quantified wands as a pair of snaps with a function from snap to snap. The encoding was already updated for non-quantified wands in https://github.com/viperproject/silicon/pull/836, which fixes an unsoundness when using the same wand in multiple ``applying``-expressions.

There is an existing implementation that also changes the encoding for quantified wands in https://github.com/viperproject/silicon/pull/849; this PR is intended to replace that existing PR, and it takes some inspiration from it and uses the tests provided by that PR.

The key problem to be solved is that, when working with quantified permissions, Silicon emits additional definitions (e.g. of snapshot maps ``sm``), that can summarize values coming from the left hand side of the wand. These definitions become part of the definition of the wand's snapshot function. So we can get a definition like 
```
forall lhs: Snap :: { mwsf_apply(lhs) } lookup(sm, r) == e(lhs) && ...
```
When the same wand is applied multiple times with different snapshots, that can lead to contradictions, because ``lookup(sm, r)`` is assumed to be equal to both ``e(lhs1)`` and ``e(lhs2)``, which might be different. The existing PR https://github.com/viperproject/silicon/pull/849 essentially tried to fix this problem retroactively, rewriting assumed declarations after the fact. This one instead uses (a slightly extended version of) Silicon's existing functionality for extending new declarations like snapshot maps with a number of quantified variables stored in the state, so we get the (unproblematic) version
```
forall lhs: Snap :: { mwsf_apply(lhs) } lookup(sm(lhs), r) == e(lhs) && ...
```
Using the existing idea of parameterizing declarations with quantified variables is more in line with the way the existing code works, and minimizes changes, which should enable us to merge some other upcoming wand-related changes more easily. 
We do, however, add a new field ``packagingWandSnapshots`` to the state instead of using the existing ``quantifiedVariables`` field because we want to treat them slightly differently and include them only when absolutely needed.

### DIFF
--- a/src/main/scala/rules/Brancher.scala
+++ b/src/main/scala/rules/Brancher.scala
@@ -50,10 +50,11 @@ object brancher extends BranchingRules {
      *   (1) the branching is due to the short-circuiting evaluation of a conjunction
      *   (2) the branch condition contains a quantified variable
      */
+    val allQuantifiedVariables = s.quantifiedVariables ++ s.packagingWandSnapshots
     val skipPathFeasibilityCheck = (
          fromShortCircuitingAnd
-      || (   s.quantifiedVariables.nonEmpty
-          && s.quantifiedVariables.map(_._1).exists(condition.freeVariables.contains))
+      || (   allQuantifiedVariables.nonEmpty
+          && allQuantifiedVariables.map(_._1).exists(condition.freeVariables.contains))
     )
 
     /* True if the then-branch is to be explored */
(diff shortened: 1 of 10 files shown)


JonasAlaif (approved):
Fix in silver:

```
//:: UnexpectedOutput(assert.failed:wand.not.found, /silicon/issue/000/)
```

Otherwise, looks great. Good job to Mr. Claude

AndreaKe (approved):
Looks good to me.

PR #986, opened 2026-06-30
refactor bounds checks for sequence indices

AndreaKe:
Refactors the assertion of sequence bounds to avoid duplicated code.

### DIFF
--- a/src/main/scala/rules/Evaluator.scala
+++ b/src/main/scala/rules/Evaluator.scala
@@ -811,18 +811,18 @@ object evaluator extends EvaluationRules {
         /* Note the reversed order of the arguments! */
 
       case ast.SeqIndex(e0, e1) =>
-				evals2(s, Seq(e0, e1), Nil, _ => pve, v)({case (s1, Seq(t0, t1), esNew, v1) =>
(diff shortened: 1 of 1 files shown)


src/main/scala/rules/Evaluator.scala
```diff
@@ -1051,6 +1001,38 @@ object evaluator extends EvaluationRules {
     resultTerm
   }
 
+
+	private def assertSeqIndexBounds(tIndex: Term, tSeq: Term, eIndex: ast.Exp, eSeq: ast.Exp, expNew: Option[Seq[ast.Exp]], pve: PartialVerificationError, s: State, v: Verifier): VerificationResult = {
+
+		def assertLtSeqLength(): VerificationResult = {
```
marcoeilers: I don't really get why this is in a nested function before the point where it is used?

marcoeilers (approved):
One question, otherwise LGTM, thanks!

Aurel300:
(Is there a lint/autoformat we can use? The PR changes spaces to tabs, which at least on GH result in misaligned indentation.)

AndreaKe:
Thanks for pointing that out. I turned the tabs into spaces and changed my IDE settings such that this doesn't happen anymore.

PR #998, opened 2026-08-11
Fix all compilation warnings and treat warnings as errors

jcp19:
Please review the unused members carefully, as it may point at implementation errors. I didn&#39;t drop dead code (e.g., unused private functions and types (e.g., `MMultiMap`)) here, but I am happy to do it too.

Regarding unused parameters, it would be good to drop them, as clients may be performing heavy computations to compute these parameters without realizing they are not ufesul

---

## Review feedback, applied

The following were settled in review and are now done:

- **`Executor.ssaifyRhs`'s `State` parameter** is dropped rather than annotated, along with its two call sites.
- **`quantifiedExp`'s `name` and `isGlobal` parameters** are dropped from the trait declaration, the implementation, the two forwarding overrides and the single call site. Its `ignore` parameter is kept and is tracked in #1006.
- **`MagicWandSnapFunctionsContributor.extractPreambleLines`** and **`DefaultStateConsolidator.partition`** are deleted rather than annotated.
- **`MMultiMap`** is deleted outright, together with the file that contained nothing else, rather than deprecated.
- **`HeapSingleton`'s simplification now matches on `v0._2`** (the value) instead of `v0._3` (the sort). It was previously a fruitless type test and therefore dead code, so this enables a simplification that had never been applied — a real change to the heap encoding, not just a warning fix. See the verification note below.
- **The function match when translating a counterexample** now rejects unexpected functions with an error message instead of an `@unchecked` scrutinee. The same match is duplicated in `Converter` and in `SiliconResolvedCounterexample`; both were changed, though only the former was flagged in review.

## Open decisions for reviewers

Everywhere the fix was not obviously behaviour-preserving, this PR keeps the existing behaviour and makes the problem explicit instead of choosing. What remains open, roughly in decreasing order of importance. Line numbers refer to the current state of the branch.

### 1. The remaining unused parameters — drop them too?

`ssaifyRhs`'s state and `quantifiedExp`'s `name`/`isGlobal` were dropped as requested. These are still annotated:

- `quantifiedExp(ignore)` (`decider/PathConditions.scala:362`) — kept deliberately: unlike the other two it is meant to have an effect, and the fact that it does not is now tracked in **#1006**.
- `Decider.debuggerAssume(de)` (`decider/Decider.scala:337`) — vestigial. Its only use was the commented-out `//TODODELETEtermSources.put(t, de)`; `termSources` no longer exists anywhere in the tree, and one of the two callers already passes `null`.
- `SiliconDebugger(identifierFactory, reporter)` (`debugger/SiliconDebugger.scala:175-176`) — the debugger uses `println` throughout and never reports.
- `quantifiedChunkSupporter.assertReadPermission(s)` (`rules/QuantifiedChunkSupport.scala:1559`).

The latter two are public signatures, which is why the PR only annotates them.

### 2. The remaining dead code — drop it too?

`extractPreambleLines`, `partition` and `MMultiMap` were deleted as requested. Still annotated `@unused`:

- `SiliconDebugger.chooseAssertion` (line 424) and `printSingleAssumption` (line 581) — interactive debugger commands that are simply not wired up, so deleting them loses the intent rather than just dead code.

### 3. Partial matches that still fail, now silently

Two matches have genuine gaps. They threw `MatchError` before this PR and still do; `@unchecked` only stops the compiler from pointing it out, so the partiality is now documented rather than fixed:

- `Converter.typecheck` (`reporting/Converter.scala:989`) does not handle `SetEntry`, `MultiSetEntry` or `PredHeapEntry`.
- `state.utils.subterms` (`state/Utils.scala:103`) does not handle `AppHint` or `MagicWandSnapshot`.

Unlike the function-translation match, which now has an error case, it is not clear here whether reaching the missing cases is a bug that should fail loudly or a case that should yield a neutral result (`false` / `Nil`), so no semantics were invented. **Decision:** which of the two?

### 4. Exhaustivity checking switched off on the `Term`/`Decl` matches

Silicon's `Term` hierarchy is deconstructed via custom flyweight extractors, so scalac cannot prove coverage even when a match is complete — this is what produced most of the "match may not be exhaustive" and "exhaustivity analysis reached max recursion depth" warnings. Where a type pattern made coverage provable, the match was rewritten that way (e.g. `IntLiteral`, `SortDecl`, `Literal`). Where it did not, the scrutinee is annotated `@unchecked`, which disables the check permanently at that site.

**Decision:** accept, or is reworking the flyweight extractors to be exhaustivity-friendly worth a follow-up?

### 5. Which warnings `-Xfatal-warnings` actually escalates

The flag makes warnings fatal but does not change which warnings are emitted. That set comes from Silver's `ThisBuild / scalacOptions`: `-deprecation -unchecked -feature -Wunused`. Categories such as most of `-Xlint`, `-Wvalue-discard` and `-Wdead-code` are not enabled and so produce nothing to be fatal about.

**Decision:** widen the emitted set in `siliconScalacOptions` as a follow-up?

Two consequences worth being aware of:

- Silicon's build is now coupled to Silver's deprecations — a Silver submodule bump that deprecates an API Silicon calls will break the build rather than warn. Escape hatch is documented in `build.sbt`: `set silicon/scalacOptions -= "-Xfatal-warnings"`.
- With `-Xfatal-warnings`, the first warning becomes an error and later warnings in the same run may not be reported, so fixing warnings can be iterative.

### 6. Behaviour change in the counterexample test helpers

In `src/test/scala/MappedCounterexampleTests.scala`, the partial functions inside `exists`/`flatMap` and `areEqualEntries` gained explicit `case _ =&gt; false` / `case _ =&gt; None` branches. Previously a non-matching shape threw `MatchError`; now it yields the neutral result. This looks like the intended semantics for a matcher, but it is a behaviour change in test infrastructure rather than a pure warning fix.

### 7. Other findings in the code merged from master

Master gained the mask-heap encoding, the greedy QP algorithm, the new counterexample format and Z3 4.16.0, none of which had been compiled with `-Xfatal-warnings`. Besides the `HeapSingleton` bug fixed above, fixing those warnings surfaced two things that are not mere annotation noise:

- **Two unreachable cases removed in `state.utils.transform`.** It matched `ZeroMask` and `PredZeroMask`, but both are `Literal`s and were therefore already handled by the preceding `case _: Literal => term`, which returns the same value the two cases returned. Removing them is behaviour-identical.
- **Silver gained the literals `RefLit` and `BackendValueLit`**, which Silicon does not encode. They are added to the lists of expressions that `evaluator.eval2` and `ExpressionTranslator` explicitly reject, matching the existing convention there, so that they fail with the intended error message instead of a `MatchError`. **Decision:** should either of them actually be supported?

### Verification

`sbt clean compile Test/compile` is warning- and error-free under the new flag, and the flag was confirmed to be fatal in all four scopes it applies to (`silicon`/`common` × `Compile`/`Test`) by injecting deliberate warnings; Silver's `scalacOptions` are unchanged and still exclude it.

Tests were run with Z3 4.16.0. Against the current state of the branch: `SiliconTests`, `SiliconTestsMaskHeapMode`, `MappedCounterexampleTests`, `GeneralCounterexampleTests` and `SiliconQuantifierWeightTests` — 1269 tests, 0 failures.

The commit that changes `HeapSingleton` was checked separately, since enabling that simplification changes the heap encoding: `SiliconTests` and `SiliconTestsQuasihavoc` (1152 tests, 0 failures), plus `SiliconTestsMaskHeapMode`, `SiliconTestsGreedyQP`, `GeneralCounterexampleTests`, `MappedCounterexampleTests`, `SiliconQuantifierWeightTests`, `TriggerGeneratorTests`, `TriggerRewriterTests` and `SimpleArithmeticTermSolverTests` (195 tests, 0 failures). The canceled and ignored tests in each run are the suites' own pre-existing known-issue annotations, and their counts did not change relative to the same suites before these modifications.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LJBLto39mSUDHwsKNhyXuN


src/main/scala/rules/Executor.scala
```diff
@@ -677,7 +677,7 @@ object executor extends ExecutionRules {
     executed
   }
 
-   private def ssaifyRhs(rhs: Term, rhsExp: ast.Exp, rhsExpNew: Option[ast.Exp], name: String, typ: ast.Type, v: Verifier, s : State): (Term, Option[ast.Exp]) = {
+   private def ssaifyRhs(rhs: Term, rhsExp: ast.Exp, rhsExpNew: Option[ast.Exp], name: String, typ: ast.Type, v: Verifier, @unused s : State): (Term, Option[ast.Exp]) = {
```
jcp19, line 680: should we drop this param altogether?

src/main/scala/supporters/MagicWandSnapFunctionsContributor.scala
```diff
@@ -62,6 +64,7 @@ class MagicWandSnapFunctionsContributor(preambleReader: PreambleReader[String, S
   override def emitAxiomsAfterAnalysis(sink: ProverLike): Unit = {}
 
   /** Helper function to transform the lines returned by the `PreambleReader`. */
+  @unused
   private def extractPreambleLines(lines: Iterable[String]*): Iterable[String] =
     lines.flatten
```
jcp19, line 69: maybe drop

src/main/scala/rules/StateConsolidator.scala
jcp19, line 345: drop?

common/src/main/scala/collections/mutable/MultiMap.scala
```diff
@@ -12,6 +12,15 @@ import scala.collection.mutable
     * sets and maps have been replaced by those that guarantee a deterministic
     * traversal order.
     */
+/* MMultiMap is unused inside Silicon, and it cannot be kept as it is: it inherits from
+ * LinkedHashMap, which is deprecated since Scala 2.13.11 because LinkedHashMap is scheduled to
+ * become final. Reimplementing MMultiMap on top of composition would change its (public) type,
+ * hence it is deprecated instead, to give potential clients outside of Silicon a chance to
+ * migrate. Deprecating MMultiMap also suppresses the warning about the deprecated inheritance,
+ * because that warning is not reported inside deprecated definitions.
+ */
```
jcp19:
```suggestion
```

src/main/scala/state/Terms.scala
```diff
@@ -2516,6 +2516,13 @@ class HeapSingleton(val at: Term, val value: Term, val sort: Sort) extends Term
...
 
 object HeapSingleton extends CondFlyweightTermFactory[(Term, Term, Sort), HeapSingleton] {
+  /* TODO: The following match is on v0._3, i.e. on the sort, which can never be a HeapLookup.
+   *       The simplification is therefore dead code and has never been applied. Presumably, the
+   *       intention was to match on v0._2, i.e. on the value. Enabling the simplification would
+   *       change how heaps are encoded, hence the behaviour is left as it is for now, and the
+   *       resulting "fruitless type test" warning is suppressed.
+   */
+  @nowarn("msg=fruitless type test")
   override def apply(v0: (Term, Term, Sort)) = v0._3 match {
```
marcoeilers: Yes, this should be ``v0._2``.

src/main/scala/rules/StateConsolidator.scala
marcoeilers: Yes

src/main/scala/reporting/Converter.scala
```diff
@@ -803,7 +804,10 @@ object Converter {
     val resSort = toSortWithSubstitutions(resTyp, s"typeError in return type $resTyp")
       .fold(err => { return errorfunc(s"$fname $err") }, identity)
 
-    val smtfunc = func match {
+    /* Only user-declared functions are translated; the built-in operations that also are
+     * ast.FuncLikes never reach this point.
+     */
+    val smtfunc = (func: @unchecked) match {
```
marcoeilers: Here we add ``@unchecked``, later we basically add an error case. I'd be in favor of also adding an error case here I think, especially since there is already a comment here that basically says what the error case should say.

src/main/scala/rules/Executor.scala
marcoeilers: Yes

src/main/scala/supporters/MagicWandSnapFunctionsContributor.scala
marcoeilers: Yes

common/src/main/scala/collections/mutable/MultiMap.scala
```diff
@@ -12,6 +12,8 @@ import scala.collection.mutable
     * sets and maps have been replaced by those that guarantee a deterministic
     * traversal order.
     */
+@deprecated("MMultiMap inherits from LinkedHashMap, which is scheduled to become final. Use a " +
+            "mutable.LinkedHashMap[A, mutable.LinkedHashSet[B]] directly instead.", "Silicon 1.1")
 trait MMultiMap[A, B] extends mutable.LinkedHashMap[A, mutable.LinkedHashSet[B]] {
```
marcoeilers: I think the whole trait might be unused; maybe we can just get rid of it instead.

src/main/scala/decider/PathConditions.scala
```diff
@@ -359,9 +361,9 @@ private trait LayeredPathConditionStackLike {
                     tQvars: Seq[Var],
                     triggers: Seq[ast.Trigger],
                     tTriggers: Seq[Trigger],
-                    name: String,
-                    isGlobal: Boolean,
-                    ignore: Term )
+                    @unused name: String,
+                    @unused isGlobal: Boolean,
+                    @unused ignore: Term )
```
marcoeilers, line 362: This does look very very suspicious. I don't know what this parameter is intended to do and if that's still required, but the only caller I can see does actually provide a value here. I'd vote to file an issue for this. I think the name and isGlobal parameters should just be dropped.

PR #1001, opened 2026-08-17
Fixing issue #996

### LINKED ISSUE #996: Silicon cannot prove resource in old context that is ensured by precondition
The attached Viper file fails to verify with Silicon because permission to `p_Global(self)` cannot be proven in the postcondition of `make_generic_Global`, even though we are in the old context and the precondition asserts this resource. The file verifies with Carbon. Note that the `m_recursive_get_last` method needs to be there for the issue to be reproducible. 

[simple.vpr.zip](https://github.com/user-attachments/files/30855383/simple.vpr.zip)

marcoeilers:
Issue #996 happens because we're calling a method whose postcondition contains an old-expression inside a ``package`` statement. The tricky thing about that is that the old-expression has to be evaluated in a heap that may never have existed inside Silicon at any point; it's the different heaps accessible inside the package statement before the call *plus* any permissions that have been consumed from outer reserve heaps while consuming the precondition.

The current code simply combines the heaps accessible inside the package statement before the call, and uses that as the old heap. This old heap simply does not contain the permissions taken from outer heaps while consuming the precondition, which is why #996 happens.

At first I thought that, since the postcondition we're supposed to produce is known to be self-framing and we only have to make sure that everything that should be available in the old heap is actually available, that we can simply combine all reserve heaps before the call, and make this combination the old heap. The problem is that this combination can be inconsistent, and then any state consolidation while producing the postcondition can lead us to assume false.

So now the solution is to (in this relative rare case) explicitly reconstruct what the old heap is, by producing the precondition again into the heap that we get after consuming it. That will give us a heap where everything taken from outer heaps is available in the wand heap. 
It's not super pretty, but I cannot think of a better solution within our current framework.

Fixes #996

### DIFF
--- a/src/main/scala/rules/Executor.scala
+++ b/src/main/scala/rules/Executor.scala
@@ -9,8 +9,6 @@ package viper.silicon.rules
 import viper.silicon.debugger.DebugExp
 import viper.silicon.common.collections.immutable.InsertionOrderedSet
 import viper.silicon.Config.JoinMode
-
-import scala.annotation.unused
 import viper.silver.cfg.silver.SilverCfg
 import viper.silver.cfg.silver.SilverCfg.{SilverBlock, SilverEdge}
 import viper.silver.verifier.{CounterexampleTransformer, NullPartialVerificationError, PartialVerificationError}
@@ -27,7 +25,7 @@ import viper.silicon.utils.ast.{BigAnd, extractPTypeFromExp, simplifyVariableNam
 import viper.silicon.utils.{freshSnap, toSf}
 import viper.silicon.verifier.Verifier
 import viper.silver.cfg.{ConditionalEdge, StatementBlock}
-
+import scala.annotation.unused
 import java.util.concurrent.atomic.AtomicInteger
 
 trait ExecutionRules extends SymbolicExecutionRules {
--- a/src/main/scala/rules/Executor.scala
+++ b/src/main/scala/rules/Executor.scala
@@ -15,7 +15,7 @@ import viper.silver.cfg.silver.SilverCfg
 import viper.silver.cfg.silver.SilverCfg.{SilverBlock, SilverEdge}
 import viper.silver.verifier.{CounterexampleTransformer, NullPartialVerificationError, PartialVerificationError}
 import viper.silver.verifier.errors._
-import viper.silver.reporter.{PathProcessedMessage, BlockReachedMessage}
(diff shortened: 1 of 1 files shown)


src/main/scala/rules/Executor.scala
```diff
@@ -24,9 +24,10 @@ import viper.silicon.logger.records.data.{CommentRecord, ConditionalEdgeRecord,
 import viper.silicon.state._
 import viper.silicon.state.terms._
 import viper.silicon.utils.ast.{BigAnd, extractPTypeFromExp, simplifyVariableName}
-import viper.silicon.utils.freshSnap
+import viper.silicon.utils.{freshSnap, toSf}
 import viper.silicon.verifier.Verifier
 import viper.silver.cfg.{ConditionalEdge, StatementBlock}
+
```
JonasAlaif: Newline needed?

src/main/scala/rules/Executor.scala
```diff
@@ -561,13 +562,49 @@ object executor extends ExecutionRules {
...
             val outs = meth.formalReturns.map(_.localVar)
             val gOuts = Store(outs.map(x => (x, v2.decider.fresh(x))).toMap)
-            val s4 = s3.copy(g = s3.g + gOuts, oldHeaps = s3.oldHeaps + (Verifier.PRE_STATE_LABEL -> magicWandSupporter.getEvalHeap(s1)))
+
+            val newOldHeaps = if (needsOldHeapReconstruct) {
+              // We're in the process of packaging a wand, which means that we cannot simply use the
+              // heap before the call as the old heap for producing the postcondition.
+              // Instead, we produce the preconditions again to get an evalHeap that contains all permissions
+              // that were consumed by the call, but does *not* contain arbitrary permissions from outer
+              // heaps. This is not pretty, but it seems necessary:
```
JonasAlaif, line 581: Is it really not possible to find the old expression stuff in the method precondition and then find that in the exhaled clientside snapshot? If that's difficult, then probably this plaster-type solution is fine

marcoeilers: There really is no way that I can think of that isn't much worse special casing. The way Silicon handles old-expressions is via a heap stored in the state under the ``old``-label; this code is constructing that heap. The easiest way to convert a snapshot that belong to a specific assertion to a heap is to produce it, which is what this code is doing.

